PR #9, opened 2020-06-22
update dependencies

mcanlas:


Baccata:
There are some new compile warnings apparently : 

[this](https://github.com/disneystreaming/weaver-test/blob/master/modules/docs/src/main/scala/weaver/Output.scala#L11) needs to be turned into a single-quoted string (`private val ansiColorRgx = "\u001b\\[([;\\d]*)m".r`) 

And some [imports](https://github.com/disneystreaming/weaver-test/blob/master/modules/scalacheck/src/weaver/scalacheck/Checkers.scala#L4-L13
) need de-duplicating :

Baccata:
Thanks for the bumps 👍

Issue #7, opened 2020-06-18
Support cross-suite resource sharing


pshirshov:
Hi,

Right now it's not possible to share one resource between several test suites. In many cases it may be a very annoying limitation.  [`distage-testkit`](https://izumi.7mind.io/distage/distage-testkit.html#testkit) collects global knowledge and builds a global test plan ahead of time which makes cross-suite shared resources possible. In case you wish we would be happy to share our experience and design ideas.

Baccata:
Hi, thanks for opening this issue. 

We are interested in providing this feature for the medium term and already have some ideas. We're aware of distage (and the izumi-reflect library that was donated to the zio org), but we're wary about not adding more dependencies to the core than we already have. 

We haven't really decided on a design just yet, but having a deeper look at how distage works has been on my todo-list for a while 😄

pshirshov:
> but having a deeper look at how distage works has been on my todo-list for a while 😄

We may setup a call and talk about these thingies if you wish :)

Baccata:
> By the way, in case you use zio you already have  

Yup, we're aware, however the core module doesn't have a dependency on zio and we intend to keep it that way.  

> "Test Server Protocol" 

That sounds interesting, I'd be keen to hear your ideas, if there's some reasonable tweaks we can make to open that a possibly, I'd definitely consider them. I'm a bit busy this week and the next, but hit me up on gitter or by email and we can have a call.

dimitarg:
Pasting something here that I drafted for our internal documentation at work

One workaround is to declare a parent suite as an object, child suites as traits, and mix in the child suites in the parent suite.
This effectively creates just one suite but lets you split that across separate files.

To put this in code, given

```scala
final case class FooResource()
final case class BarResource(value: Int)
final case class Resources(fooResource: FooResource, barResource: BarResource)
```

, declare "child" suite traits

```scala
import weaver.IOSuite

trait ChildSuite extends IOSuite { self: AllSuites.type =>
  test("accessing shared resource") { r =>
    expect(r.barResource.value == 42)
  }
}
```

and a "parent" suite instantiating the children

```scala
import cats.effect.{IO, Resource}
import weaver.IOSuite

object AllSuites extends IOSuite
  with ChildSuite {
  override type Res = Resources

  override def sharedResource: Resource[IO, Resources] = for {
    foo <- Resource.pure(FooResource())
    bar <- Resource.pure(BarResource(42))
[...]

PR #26, opened 2020-06-25
Cross-suite resource sharing

Baccata:
Resolves https://github.com/disneystreaming/weaver-test/issues/7

* When running suites in the context of a build tool, it is currently
impossible to share resources (http clients, databases) between several
suites.

This PR offers a non-intrusive, source-compatible way of doing it.

1. Users can implement `weaver.framework.GlobalResourcesInit` as
custom objects alongside their test suites. This interface is detected
as a task by build tools and managed by our framework implementation.
The method users have to implement provides a "store" (a shared state)
that they can use to store arbitrary objects in a cats.effect.Resource
program. That cats.effect.Resource is used by the framework to allocate
and clean-up resources the users might want to share between tests.
2. What the user can store is expressed in terms of a `ResourceTag`
that needs to be passed implicitly, with a default implementation based
on ClassTags and therefore subject to type-erasure. We try and protect
against type-erasure by preventing the use of this tag for types of kind
`* -> *` or `(* -> *) -> *`.
3. Users can then declare a constructor to their suite
(as opposed to having them be a static object). The constructor must
take a single `weaver.framework.GlobalResources` parameter. This object
offers a read-only interface to the shared store.

Shared resources get cleaned up after all the suites have run.

### DIFF
--- a/modules/framework/src/weaver/framework/TestFramework.scala
+++ b/modules/framework/src/weaver/framework/TestFramework.scala
@@ -8,7 +8,8 @@ class TestFramework extends BaseFramework {
   def name(): String = "weaver"
 
   def fingerprints(): Array[Fingerprint] = {
-    Array(TestFramework.ModuleFingerprint)
+    Array(TestFramework.GlobalResourcesFingerprint,
+          TestFramework.ModuleFingerprint)
   }
 
   def runner(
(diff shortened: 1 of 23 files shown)


modules/framework/src/weaver/framework/TestFramework.scala
```diff
@@ -33,9 +39,42 @@ object TestFramework {
...
+   */
+  object GlobalResourcesSharingFingerprint extends WeaverFingerprint {
+    val isModule                           = false
+    def requireNoArgConstructor(): Boolean = false
+    def superclassName(): String           = "weaver.BaseSuiteClass"
+  }
+
+  /**
+   * A fingerprint that searches only for singleton objects
+   * of type [[weaver.EffectSuite]].
```
Baccata, line 60: Change this

PR #33, opened 2020-07-03
Use scalafix

Slakah:
Hi folks,

What this does:

## Interesting

* Enables [scalafix](https://scalacenter.github.io/scalafix/)
* Organise imports using [scalafix-organize-imports](https://github.com/liancheng/scalafix-organize-imports)
* Add a sbt `fix` alias to run scalafix and scalafmt
* Run `fix` over the project and commit the results

## Refactoring

* In SBT files prefer the syntax `<scope> / <scope>` over `<scope> in <scope>`
* Alias task list are now constructed from a `Seq` - makes it easier to scan

### DIFF
--- a/.scalafix.conf
+++ b/.scalafix.conf
@@ -21,3 +21,13 @@ DisableSyntax.noFinalize = true
 DisableSyntax.noValPatterns = false
 DisableSyntax.noUniversalEquality = false
 
+OrganizeImports {
+  coalesceToWildcardImportThreshold = 2147483647 # Int.MaxValue
+  expandRelative = false
+  groupExplicitlyImportedImplicitsSeparately = false
+  groupedImports = Merge
+  groups = ["re:javax?\\.", "scala.", "*"]
+  importSelectorsOrder = Ascii
+  importsOrder = Ascii
+  removeUnused = true
+}
--- a/.scalafix.conf
+++ b/.scalafix.conf
@@ -7,16 +7,16 @@ rules = [
   ProcedureSyntax
 ]
 
-DisableSyntax.noVars = true
-DisableSyntax.noThrows = true
-DisableSyntax.noNulls = true
+DisableSyntax.noVars = false
+DisableSyntax.noThrows = false
+DisableSyntax.noNulls = false
 DisableSyntax.noReturns = true
 DisableSyntax.noWhileLoops = true
 DisableSyntax.noAsInstanceOf = false
-DisableSyntax.noIsInstanceOf = true
+DisableSyntax.noIsInstanceOf = false
 DisableSyntax.noXml = true
 DisableSyntax.noDefaultArgs = false
-DisableSyntax.noFinalVal = true
+DisableSyntax.noFinalVal = false
 DisableSyntax.noFinalize = true
 DisableSyntax.noValPatterns = false
 DisableSyntax.noUniversalEquality = false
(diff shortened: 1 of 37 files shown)


.scalafix.conf
```diff
@@ -0,0 +1,23 @@
+rules = [
+  DisableSyntax
+  LeakingImplicitClassVal
+  NoAutoTupling
+  NoValInForComprehension
+  OrganizeImports
+  ProcedureSyntax
+]
+
+DisableSyntax.noVars = true
```
Baccata, line 10:
```suggestion
```

.scalafix.conf
```diff
@@ -0,0 +1,23 @@
...
+DisableSyntax.noThrows = true
```
Baccata, line 11:
```suggestion
```

.scalafix.conf
```diff
@@ -0,0 +1,23 @@
...
+DisableSyntax.noNulls = true
```
Baccata, line 12:
```suggestion
```

.scalafix.conf
```diff
@@ -0,0 +1,23 @@
...
+DisableSyntax.noReturns = true
+DisableSyntax.noWhileLoops = true
+DisableSyntax.noAsInstanceOf = false
+DisableSyntax.noIsInstanceOf = true
```
Baccata, line 16:
```suggestion
```

Baccata:
> Use Seq + .mkString to reduce the line length of this line 

good idea

> Are the current rules acceptable? The diff after applying them would be this, this means we would end up with a few @SuppressWarnings annotations 

Added code suggestions for the rules I don't want. 

> Add a fix alias, to run scalafmt/scalafix across the project 

Cool :+1: 

> Replace sbt syntactic sugar like scalafixDependencies in ThisBuild with ThisBuild / scalafixDependencies in the build.sbt file

I'm fine with this, but I'd like for the same syntax to be used consistently across the build definition (and whatever's in /project)

modules/framework/test/src/TestConsole.scala
```diff
@@ -3,7 +3,7 @@ package framework
 package test
 
 object TestConsole {
-  private val ansiColorRgx = """\u001b\[([;\d]*)m""".r
+  private val ansiColorRgx = "\u001b\\[([;\\d]*)m".r
```
Slakah, line 6: Dunno whether this is actually required, but the warning was bugging me.

Slakah:
@Baccata , this should be good to go now. Let me know if you want anything else changing.

.scalafix.conf
```diff
@@ -0,0 +1,33 @@
...
+DisableSyntax.noFinalize = true
+DisableSyntax.noValPatterns = false
+DisableSyntax.noUniversalEquality = false
+
+OrganizeImports {
+  coalesceToWildcardImportThreshold = 2147483647 # Int.MaxValue
+  expandRelative = false
+  groupExplicitlyImportedImplicitsSeparately = false
+  groupedImports = Merge
+  groups = ["re:javax?\\.", "scala.", "*"]
```
Baccata, line 29: Can we add a couple groups here ? 

```suggestion
  groups = ["re:javax?\\.", "scala.", "cats.", "weaver.", "*"]
```

Baccata:
Thanks a lot for this :+1:

PR #37, opened 2020-07-07
Add an interop module for specs2

mattroberts297:
So that people can more easily migrate from specs2 whilst still
leveraging the matchers provided by that framework.

**TODO**

- [x] Fix up JS project
- [x] Rename packages
- [x] Remove accidental inclusion of bloop plugin
- [x] Add documentation
- [x] Investigate build OOM issue

### DIFF
--- a/build.sbt
+++ b/build.sbt
@@ -46,10 +46,12 @@ lazy val root = project
              frameworkJVM,
              scalacheckJVM,
              zioJVM,
+             specs2JVM,
              coreJS,
              frameworkJS,
              scalacheckJS,
-             zioJS)
+             zioJS,
+             specs2JS)
   .configure(WeaverPlugin.profile)
   .settings(WeaverPlugin.doNotPublishArtifact)
   .settings(
@@ -73,6 +75,11 @@ lazy val core = crossProject(JSPlatform, JVMPlatform)
   .jvmSettings(
     libraryDependencies ++= Seq(
       "org.scala-js" %%% "scalajs-stubs" % "1.0.0" % "provided"
+    )
+  )
+  .jsSettings(
+    libraryDependencies ++= Seq(
+      "io.github.cquiroz" %%% "scala-java-time" % "2.0.0"
     )
   )
 
@@ -133,14 +140,30 @@ lazy val scalacheck = crossProject(JSPlatform, JVMPlatform)
   .settings(WeaverPlugin.simpleLayout)
   .settings(
     libraryDependencies ++= Seq(
-      "org.scalacheck"    %%% "scalacheck"      % "1.14.3",
-      "io.github.cquiroz" %%% "scala-java-time" % "2.0.0" % Test
+      "org.scalacheck" %%% "scalacheck" % "1.14.3"
     ),
     scalaJSLinkerConfig ~= { _.withModuleKind(ModuleKind.CommonJSModule) }
   )
 
 lazy val scalacheckJVM = scalacheck.jvm
 lazy val scalacheckJS  = scalacheck.js
+
+lazy val specs2 = crossProject(JSPlatform, JVMPlatform)
+  .crossType(CrossType.Pure)
+  .in(file("modules/specs2"))
+  .dependsOn(core, framework % "test->compile")
+  .configure(WeaverPlugin.profile)
+  .settings(
+    name := "specs2",
+    libraryDependencies ++= Seq(
+      "org.specs2" %%% "specs2-matcher" % "4.10.0"
+    ),
+    scalaJSLinkerConfig ~= { _.withModuleKind(ModuleKind.CommonJSModule) }
+  )
+  .settings(WeaverPlugin.simpleLayout)
+
+lazy val specs2JVM = specs2.jvm
+lazy val specs2JS  = specs2.js
 
 lazy val zio = crossProject(JSPlatform, JVMPlatform)
   .crossType(CrossType.Pure)
--- a/build.sbt
+++ b/build.sbt
@@ -89,7 +89,7 @@ lazy val coreJS  = core.js
 lazy val docs = project
   .in(file("modules/docs"))
   .enablePlugins(DocusaurusPlugin, MdocPlugin)
-  .dependsOn(coreJVM, frameworkJVM, scalacheckJVM, zioJVM)
+  .dependsOn(coreJVM, frameworkJVM, scalacheckJVM, zioJVM, specs2JVM)
   .settings(
     moduleName := "docs",
     watchSources += (ThisBuild / baseDirectory).value / "docs",
--- a/project/plugins.sbt
+++ b/project/plugins.sbt
@@ -12,4 +12,3 @@ addSbtPlugin("org.scalameta"        % "sbt-scalafmt"                  % "2.3.0")
 addSbtPlugin("org.scalameta"        % "sbt-scalafmt"                  % "2.4.2")
 addSbtPlugin("org.scoverage"        % "sbt-scoverage"                 % "1.6.1")
 addSbtPlugin("org.scalameta"        % "sbt-mdoc"                      % "2.2.5")
-
--- a/.drone.yml
+++ b/.drone.yml
@@ -12,6 +12,7 @@ steps:
   environment:
     COURSIER_CACHE: .cache/coursier
     JAVA_OPTS: "-XX:MaxMetaspaceSize=1g -Xms1g -Xmx2g -Xss2M"
+    SBT_OPTS: "-XX:+CMSClassUnloadingEnabled -XX:MaxMetaspaceSize=2g -Xmx2g"
   volumes:
     - name: sbt-cache
       path: /root/.sbt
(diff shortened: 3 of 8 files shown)


modules/interop-specs2/test/src/weaver/interop/specs2/MatchersSpec.scala
```diff
@@ -0,0 +1,25 @@
+package weaver.interop.specs2
+
+import weaver.SimpleIOSuite
+
+object MatchersSpec extends SimpleIOSuite with IOMatchers {
```
Baccata, line 5: There's a [DogFood](https://github.com/disneystreaming/weaver-test/blob/master/modules/framework/src/weaver/framework/DogFood.scala) trait that contains the necessary functions to manually run the framework (it's very meta) and intercept the events it fires as well as the logs, which could be used to test the negative cases, but it involves a fair amount of string matching. 

[Example usage](https://github.com/disneystreaming/weaver-test/blob/master/modules/framework/test/src/DogFoodTests.scala#L78) 

Alternatively, you could tap into the [spec](https://github.com/disneystreaming/weaver-test/blob/master/modules/core/src/weaver/suites.scala#L24) method (nesting a suite containing failing tests inside another object to prevent the actual framework instance from detecting it). 

That'd let you test that conversions are actually capturing the correct sourcecode metadata.

modules/interop-specs2/src/weaver/interop/specs2/Matchers.scala
```diff
@@ -0,0 +1,28 @@
...
+import cats.data.Validated
+import cats.effect.IO
+import weaver.EffectSuite
+import org.specs2.matcher.MustMatchers
+import org.specs2.matcher.MatchResult
+import weaver.AssertionException
+import weaver.Expectations
+import weaver.SourceLocation
+
+trait Matchers[F[_]] extends MustMatchers {
```
Baccata, line 13: that's pretty cool, I had no idea `specs2` was serving its matchers in an independent artifact, nice find :+1:

modules/interop-specs2/test/src/weaver/interop/specs2/MatchersSpec.scala
```diff
@@ -0,0 +1,25 @@
+package weaver.interop.specs2
```
Baccata, line 1: I'm waging a war against empty packages, I think it's an arbitrary javaism that makes apis genuinely harder to use, by making the user having to refer themselves to the documentation more often. That's (partly) why the top package is `weaver` as opposed to `com.disneystreaming.weaver`

I think the following is fine : 

```scala
import weaver.specs2 
```

If you're concerned about package collision (which can be a genuine concern depending on the UX of the library you're trying to interop with), I think `weaver.specs2compat` would do (the zio package is `ziocompat` for this reason). It's auto-completion friendly.

build.sbt
```diff
@@ -112,6 +114,23 @@ lazy val scalacheck = crossProject(JSPlatform, JVMPlatform)
 lazy val scalacheckJVM = scalacheck.jvm
 lazy val scalacheckJS  = scalacheck.js
 
+lazy val interopSpecs2 = crossProject(JSPlatform, JVMPlatform)
```
Baccata, line 117: I'd call it `specs2`, making the artifact's name `weaver-specs2`, which is consistent with the rest of the interop packages (zio / scalacheck)

Baccata:
Here's a patch that fixes the build. Apologies for that, we only recently re-enabled scala-js support and the necessary dependency had been luckily added to the downstream artifacts' test scopes. 

``` 
From 5c4dc58d84f74a7aa77ac14c7de06dd05310f00e Mon Sep 17 00:00:00 2001
From: =?UTF-8?q?Olivier=20M=C3=A9lois?= <olivier.melois@disneystreaming.com>
Date: Tue, 7 Jul 2020 23:58:44 +0200
Subject: [PATCH] Fix sjs tests + run format

---
 build.sbt                                             | 11 +++++++----
 .../src/weaver/interop/specs2/Matchers.scala          |  7 +++++--
 2 files changed, 12 insertions(+), 6 deletions(-)

diff --git a/build.sbt b/build.sbt
index 1f63952..6320890 100644
--- a/build.sbt
+++ b/build.sbt
@@ -47,6 +47,11 @@ lazy val core = crossProject(JSPlatform, JVMPlatform)
       "org.scala-js" %%% "scalajs-stubs" % "1.0.0" % "provided"
     )
   )
+  .jsSettings(
+    libraryDependencies ++= Seq(
+      "io.github.cquiroz" %%% "scala-java-time" % "2.0.0"
+    )
+  )
 
 lazy val coreJVM = core.jvm
 lazy val coreJS  = core.js
@@ -105,8 +110,7 @@ lazy val scalacheck = crossProject(JSPlatform, JVMPlatform)
   .settings(WeaverPlugin.simpleLayout)
   .settings(
     libraryDependencies ++= Seq(
-      "org.scalacheck"    %%% "scalacheck"      % "1.14.3",
-      "io.github.cquiroz" %%% "scala-java-time" % "2.0.0" % Test
+      "org.scalacheck"    %%% "scalacheck"      % "1.14.3"
     ),
     scalaJSLinkerConfig ~= { _.withModuleKind(ModuleKind.CommonJSModule) }
   )
@@ -139,8 +143,7 @@ lazy val zio = crossProject(JSPlatform, JVMPlatform)
   .settings(WeaverPlugin.simpleLayout)
   .settings(
     libraryDependencies ++= Seq(
-      "dev.zio"           %%% "zio-interop-cats" % "2.1.3.0-RC16",
-      "io.github.cquiroz" %%% "scala-java-time"  % "2.0.0"
+      "dev.zio"           %%% "zio-interop-cats" % "2.1.3.0-RC16"
     ),
     scalaJSLinkerConfig ~= { _.withModuleKind(ModuleKind.CommonJSModule) }
   )
diff --git a/modules/interop-specs2/src/weaver/interop/specs2/Matchers.scala b/modules/interop-specs2/src/weaver/interop/specs2/Matchers.scala
index 65f0ca0..569ab39 100644
--- a/modules/interop-specs2/src/weaver/interop/specs2/Matchers.scala
+++ b/modules/interop-specs2/src/weaver/interop/specs2/Matchers.scala
@@ -13,11 +13,14 @@ import weaver.SourceLocation
 trait Matchers[F[_]] extends MustMatchers {
   self: EffectSuite[F] =>
 
-  implicit def toExpectations[A](m: MatchResult[A])(implicit pos: SourceLocation): F[Expectations] = effect.pure {
+  implicit def toExpectations[A](m: MatchResult[A])(implicit
+  pos: SourceLocation): F[Expectations] = effect.pure {
     if (m.toResult.isSuccess) {
       Monoid[Expectations].empty
     } else {
-      Expectations(Validated.invalidNel(new AssertionException(m.toResult.message, pos)))
+      Expectations(Validated.invalidNel(new AssertionException(
+        m.toResult.message,
+        pos)))
     }
   }
 
-- 
2.22.0
```

project/plugins.sbt
```diff
@@ -12,4 +12,4 @@ addSbtPlugin("org.scalameta"        % "sbt-scalafmt"                  % "2.3.0")
 addSbtPlugin("org.scalameta"        % "sbt-scalafmt"                  % "2.4.0")
 addSbtPlugin("org.scoverage"        % "sbt-scoverage"                 % "1.6.1")
 addSbtPlugin("org.scalameta"        % "sbt-mdoc"                      % "2.2.3")
-
+addSbtPlugin("ch.epfl.scala"        % "sbt-bloop"                     % "1.4.3")
```
Baccata, line 15:
```suggestion
```

Please add this to your [global settings](https://www.scala-sbt.org/1.x/docs/Global-Settings.html#Global+Settings+using+a+Global+Plugin) rather than here. You can probably even retrieve the version dynamically by shelling out to `bloop --help` and parsing the first line, so that it always matches the version you have installed.

Baccata:
The code looks good. Few points however : 

* Build is failing because of an OOM. I'm failing to find the spec of the machines the CI runs on, but we can try to hardcode an arbitrary Xmx value [here](https://github.com/disneystreaming/weaver-test/blob/master/.drone.yml#L12). I personally use the following in my .bash_profile : `export SBT_OPTS="-Xms1024M -Xmx1024M -Xss2M"`
* could you remove the [bloop plugin](https://github.com/disneystreaming/weaver-test/pull/37#pullrequestreview-445319818) ? If it's not used on CI, it doesn't have a place in the build definition 
* Can you add a little [documentation](https://github.com/disneystreaming/weaver-test/tree/master/docs) ? You'll need to amend [this](https://github.com/disneystreaming/weaver-test/blob/master/website/sidebars.json#L7). You can build the website locally via : `docs/docusaurusCreateSite` in sbt, then `cd website && yarn start`

project/plugins.sbt
mattroberts297: Not intentional - sorry. Slipped in whilst (lazily) rebasing!

Baccata:
Thanks for the contribution 😸

PR #52, opened 2020-08-15
Improve documentation; upgrade Scala

mcanlas:
Quality of life stuff, in patches


project/WeaverPlugin.scala
```diff
@@ -90,7 +90,6 @@ object WeaverPlugin extends AutoPlugin {
     "-Xlint:inaccessible",           // Warn about inaccessible types in method signatures.
     "-Xlint:infer-any",              // Warn when a type argument is inferred to be `Any`.
     "-Xlint:missing-interpolator",   // A string literal appears to be missing an interpolator id.
-    "-Xlint:nullary-override",       // Warn when non-nullary `def f()' overrides nullary `def f'.
```
Baccata, line 93: I see this was removed in 2.13.3, was it also removed in 2.12.12 ? If not, could you move it to the 2.12_only list of compiler options rather than removing altogether ?

README.md
```diff
@@ -49,6 +49,13 @@ which generally keeps tests cleaner / clearer.
 
 ## API
 
+Weaver provides two types of testing suites.
```
Baccata, line 52: Slightly inaccurate : there are more than two (in particular if you include zio). Not sure how to phrase it whilst not getting into details. Maybe a "mainly" dropped in there would suffice

project/WeaverPlugin.scala
mcanlas: Moved in force push.

README.md
mcanlas: I found the `zio` stuff and added a new section to clarify (and clarified the original wording).

Baccata: Perfect, thanks !

PR #56, opened 2020-08-19
Add tagging documentation

marcraminv:


docs/tagging.md
```diff
@@ -0,0 +1,32 @@
+---
+id: tagging
+title: Tagging
+---
+
+Weaver brings some default tags as `ignore` or `cancel` to disable the tests temporally.
```
Baccata, line 6:
```suggestion
Weaver provides some constructs to dynamically tag tests as `ignored` or `cancelled` :   
```

marcraminv: thanks for feedback @Baccata !

docs/tagging.md
```diff
@@ -0,0 +1,32 @@
...
+
+object MySuite extends SimpleIOSuite {
+
+
+  simpleTest("hello tag ignore") {
+    for {
+      x <- IO.delay(1)
+      y <- IO.delay(2)
+    } yield expect(x == y)
+    ignore("I want to ignore this test because is flaky")
```
Baccata, line 20: though this works, the for-comprehension is not monadically tied with the `ignore` call. A more relevant example would be something like : 

```scala
import cats.implicits._ 

simpleTest("only on CI") {
    for {
      onCI <- IO(sys.env.get("CI").isDefined)
      _    <- ignore("not on CI").unlessA(onCI)
      x    <- IO.delay(1)
      y    <- IO.delay(2)
    } yield expect(x == y)
}
```

Same goes for cancel (if you want to vary the example, you could cancel the test when it's the morning or something)

marcraminv: Right, has sense inside the for-comprehension, I change it. 👍

Baccata (approved):
Thanks for contributing 😄

PR #59, opened 2020-08-27
Adds a command-line interface module

Baccata:
The main goal of this is to offer a way for Intellij to communicate with weaver, even if there's a mismatch of version between the IDE and the test framework 

**/!\ contains a huge json file**

### DIFF
--- a/modules/cli/src/weaver/cli/package.scala
+++ b/modules/cli/src/weaver/cli/package.scala
@@ -1,6 +1,0 @@
-package weaver
-
-
-package object cli {
-
-}
--- a/modules/cli/src/weaver/cli/package.scala
+++ b/modules/cli/src/weaver/cli/package.scala
@@ -1,0 +1,6 @@
+package weaver
+
+
+package object cli {
+
+}
--- a/modules/codecs/test/resources/golden.json
+++ b/modules/codecs/test/resources/golden.json
@@ -1,0 +1,1459 @@
+{
+  "outcome": {
+    "name": "outcome",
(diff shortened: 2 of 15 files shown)


modules/core/src/weaver/Log.scala
```diff
@@ -54,13 +54,20 @@ object Log {
...
+  case object warn  extends Level("warn")
+  case object debug extends Level("debug")
+  case object error extends Level("error")
 
   object Level {
+    val values: List[Level] = List(info, warn, debug, error)
+
+    def fromString(s: String): Either[String, Level] =
+      values.find(_.label == s) match {
+        case other => Left(s"$other is not a valid log-level")
```
keynmol, line 68: I'm confused - should there be a successful branch?

keynmol:
```suggestion
        case None            => Left(s"$other is not a valid log-level")
        case Some(level) => Right(level)
```

modules/cli/src/weaver/cli/package.scala
```diff
@@ -0,0 +1,6 @@
+package weaver
+
+
+package object cli {
```
keynmol, line 4: is this needed?

modules/codecs/test/resources/golden.json
```diff
@@ -0,0 +1,1459 @@
+{
```
Baccata, line 1: This is used in a golden test, in order to prevent accidental breakage of protocol

Issue #43, opened 2020-07-29
Add monix support 


Baccata:
Pretty self-explanatory. A monix module should be added and contain monix-specific suite traits setting the effect type to `monix.eval.Task` and have sensible defaults that caters to Monix's runtime.

PR #63, opened 2020-09-03
Monix support

bfdes:
Attempt to address https://github.com/disneystreaming/weaver-test/issues/43.

### DIFF
--- a/README.md
+++ b/README.md
@@ -12,7 +12,7 @@
 
 
 A test-framework built on [cats-effect](https://github.com/typelevel/cats-effect) and
-[fs2](https://github.com/functional-streams-for-scala/fs2), with [zio](https://zio.dev) interop.
+[fs2](https://github.com/functional-streams-for-scala/fs2), with [zio](https://zio.dev) and [monix](https://monix.io) interop.
 
 ## Installation
 
@@ -28,6 +28,9 @@ testFrameworks += new TestFramework("weaver.framework.TestFramework")
 
 // optionally (for ZIO usage)
 libraryDependencies +=  "com.disneystreaming" %% "weaver-zio" % "x.y.z" % Test
+
+// optionally (for Monix usage)
+libraryDependencies +=  "com.disneystreaming" %% "weaver-monix" % "x.y.z" % Test
 
 // optionally (for Scalacheck usage)
 libraryDependencies +=  "com.disneystreaming" %% "weaver-scalacheck" % "x.y.z" % Test
@@ -131,14 +134,18 @@ object MySuite extends IOSuite {
 
 #### Other suites
 
-Weaver also includes support for `ZIO`-based suites via the optional `weaver-zio` dependency.
+Weaver also includes support for 
+* `ZIO`-based suites via the optional `weaver-zio` dependency
+* `Monix`-based suites via the optional `weaver-monix` dependency
 
 | Alias | Suite name | Provided by | Use case |
 | --- | --- | --- | --- |
-| `SimpleIOSuite`  | `SimpleMutableIOSuite`       | `weaver-framework` | Each test is a standalone `IO` action
-| `IOSuite`        | `MutableIOSuite`             | `weaver-framework` | Each test needs access to a shared `Resource`
-| `SimpleZIOSuite` | `SimpleMutableZIOSuite`      | `weaver-zio`       | Each test is a standalone `ZIO` action
-| `ZIOSuite[R]`    | `MutableZIOSuite[R]`         | `weaver-zio`       | Each test needs access to a shared `ZLayer`
+| `SimpleIOSuite`   | `SimpleMutableIOSuite`       | `weaver-framework` | Each test is a standalone `IO` action
+| `IOSuite`         | `MutableIOSuite`             | `weaver-framework` | Each test needs access to a shared `Resource`
+| `SimpleZIOSuite`  | `SimpleMutableZIOSuite`      | `weaver-zio`       | Each test is a standalone `ZIO` action
+| `ZIOSuite[R]`     | `MutableZIOSuite[R]`         | `weaver-zio`       | Each test needs access to a shared `ZLayer`
+| `SimpleMonixSuite`| `SimpleMutableMonixSuite`    | `weaver-monix`     | Each test is a standalone `Task` action
+| `MonixSuite`      | `MutableMonixSuite`          | `weaver-monix`     | Each test needs access to a shared `Resource`
 
 ### Expectations (assertions)
 
--- a/modules/monix/src/weaver/monixcompat/suites.scala
+++ b/modules/monix/src/weaver/monixcompat/suites.scala
@@ -8,12 +8,12 @@ import monix.eval.instances._
 import monix.execution.Scheduler
 import cats.effect.{ Timer, ContextShift, ConcurrentEffect }
 
-trait BaseMonixSuite { self : ConcurrentEffectSuite[Task] =>
-  val ec = scala.concurrent.ExecutionContext.global
-  val scheduler = Scheduler(ec)
-  implicit def timer : Timer[Task] = Task.timer(scheduler)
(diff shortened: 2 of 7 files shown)


modules/monix/src/weaver/monixcompat/suites.scala
```diff
@@ -0,0 +1,24 @@
...
+import monix.eval.instances._
+import monix.execution.Scheduler
+
+trait BaseMonixSuite { self: ConcurrentEffectSuite[Task] =>
+  val ec                              = scala.concurrent.ExecutionContext.global
+  val scheduler                       = Scheduler(ec)
+  implicit def timer: Timer[Task]     = Task.timer(scheduler)
+  implicit def cs: ContextShift[Task] = Task.contextShift(scheduler)
+  implicit def effect: ConcurrentEffect[Task] =
+    new CatsConcurrentEffectForTask()(scheduler, Task.defaultOptions)
```
bfdes, line 16: It remains to see whether these defaults are sensible.

modules/monix/src/weaver/monixcompat/suites.scala
```diff
@@ -0,0 +1,24 @@
...
+}
+
+trait MutableMonixSuite extends MutableFSuite[Task] with BaseMonixSuite
+
+trait SimpleMutableMonixSuite extends MutableMonixSuite {
+  type Res = Unit
+  def sharedResource: Resource[Task, Unit] = Resource.pure(())
```
bfdes, line 23: Still looking into whether `Bracket` should be used instead of `Resource`. From my understanding, Monix `Task.bracket` simply creates a Cats Effect `Bracket`...

modules/monix/test/src/weaver/monixcompat/MonixSuiteTest.scala
```diff
@@ -0,0 +1,63 @@
...
+import cats.effect.Resource
+
+import weaver.framework.DogFood
+
+import monix.eval.Task
+import sbt.testing.Status
+
+object MonixSuiteTest extends MonixSuite {
+  type Res = Unit
+  override def sharedResource: Resource[Task, Unit] = Resource.pure(())
```
bfdes, line 12: TODO: Test resource access patterns as in ZIO combat tests.

modules/monix/src/weaver/monixcompat/suites.scala
```diff
@@ -0,0 +1,24 @@
...
+package monixcompat
+
+import cats.effect.{ ConcurrentEffect, ContextShift, Resource, Timer }
+
+import monix.eval.Task
+import monix.eval.instances._
+import monix.execution.Scheduler
+
+trait BaseMonixSuite { self: ConcurrentEffectSuite[Task] =>
+  val ec                              = scala.concurrent.ExecutionContext.global
```
Baccata, line 11: I think I wouldn't expose this one publicly. `Scheduler` is more central to monix, and also extends ExecutionContext

modules/monix/src/weaver/monixcompat/suites.scala
Baccata: I think it's okay

modules/monix/src/weaver/monixcompat/suites.scala
Baccata: Let's leave it as Resource. Monix doesn't have (afaik) a competing construct to `cats.effect.Resource`(zio however does). `Resource` makes use of the `Bracket` typeclass, which is implemented using `Task#bracket` in this case

modules/monix/test/src/weaver/monixcompat/MonixSuiteTest.scala
Baccata: Not necessary 😄 , you're extending `MutableFSuite` which is already tested. zio has specific tests because it extends an interface that's higher in the hierarchy, which does not make use of `cats.effect.Resource`

README.md
```diff
@@ -131,14 +134,18 @@ object MySuite extends IOSuite {
 
 #### Other suites
 
-Weaver also includes support for `ZIO`-based suites via the optional `weaver-zio` dependency.
+Weaver also includes support for
```
Baccata, line 137: Thanks for the added bits in the readme. Could you also add a page in the website ? You'll have to : 

* add a new markdown file like [this one]((https://github.com/bfdes/weaver-test/blob/monix-support/docs/zio_usage.md) in the same directory
* reference it  [there](https://github.com/bfdes/weaver-test/blob/master/website/sidebars.json#L15 )

You can build the website locally following these [instructions](https://scalameta.org/mdoc/docs/docusaurus.html#create-static-site) and call `yarn run start` to automatically open it in your browser

modules/monix/src/weaver/monixcompat/suites.scala
```diff
@@ -0,0 +1,24 @@
...
+  val scheduler                       = Scheduler(ec)
+  implicit def timer: Timer[Task]     = Task.timer(scheduler)
+  implicit def cs: ContextShift[Task] = Task.contextShift(scheduler)
+  implicit def effect: ConcurrentEffect[Task] =
+    new CatsConcurrentEffectForTask()(scheduler, Task.defaultOptions)
+}
+
+trait MutableMonixSuite extends MutableFSuite[Task] with BaseMonixSuite
```
Baccata, line 19: I'd call these `TaskSuite` for coherence, as the name of the effect type is `Task` (ie the cats-effect specific one is called `IOSuite` as opposed to `CatsSuite`)

Baccata:
Hi @bfdes , thanks a lot for this 😄 . I think it's closer to ready than you might think, just need a page added to the website, and a couple modifications in names and accessors

modules/monix/src/weaver/monixcompat/suites.scala
bfdes: No problem. I was a bit confused as the ZIO suite is called `MutableZIOSuite`, and I thought the ZIO effect type was also called `Task`?

modules/monix/src/weaver/monixcompat/suites.scala
bfdes: 👍 I'll replace it with `import monix.execution.Scheduler.Implicits.global`, which I imagine wraps `scala.concurrent.ExecutionContext.global`.

Baccata: I'd just do `val scheduler : Scheduler = monix.execution.Scheduler.global`, no need to refer to it implicitly I think, considering it's not used in combinator methods on Task

modules/monix/src/weaver/monixcompat/suites.scala
Baccata: You're not wrong, but `Task[R, A]` is type alias for `ZIO[R, Throwable, A]`, so the actual effect type is called `ZIO`

modules/monix/test/src/weaver/monixcompat/MonixSuiteTest.scala
bfdes: I found it a bit hard to follow the base framework tests, so I wasn't sure. 
In that case, I can replace `MonixSuite` here with `SimpleMonixSuite` here. Since that is essentially what it is.

modules/monix/src/weaver/monixcompat/suites.scala
bfdes: Addressed in b644c8a823f102f8bcc638299adb09818008ed9c.

modules/monix/src/weaver/monixcompat/suites.scala
bfdes: Addressed in b644c8a823f102f8bcc638299adb09818008ed9c.

README.md
bfdes: Addressed in 4b29674bf0724234761a5a08f6337557ec19de7f.

Baccata (approved):

Baccata:
Thanks a lot for this @bfdes !

PR #64, opened 2020-09-07
Decouple EffectSuite from Expecations.Helpers

Baccata:
Makes it possible to tap into some suite abstract classes / traits
without buying into the default ways of constructing expectations.

This change should retain compatibility in most cases for end users.

### DIFF
--- a/modules/core/src/weaver/Expectations.scala
+++ b/modules/core/src/weaver/Expectations.scala
@@ -83,6 +83,12 @@ object Expectations {
 
   trait Helpers {
 
+    /**
+     * Expect macros
+     */
+    def expect = new Expect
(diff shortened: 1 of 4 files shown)


modules/core/src/weaver/Expectations.scala
```diff
@@ -83,6 +83,12 @@ object Expectations {
 
   trait Helpers {
 
+    /**
+     * Expect macros
+     */
+    def expect = new Expect
+    def assert = new Expect
```
keynmol, line 90: Nice - I've fallen into a pattern of re-defining those in custom traits that are not part of the suite

keynmol (approved):

PR #74, opened 2020-09-11
Adds an intellij plugin

Baccata:
**This change is breaking compatibility**, though most users should not really see it. 

Adds a plugin that lets users run tests from Intellij, by clicking on buttons on the LHS of the editor, that appear on the same line as suite declarations **OR** tests

Screenshot 2020-09-11 at 12 27 32

* The main change in weaver-core is that `test` methods receive a `TestName` as opposed to a string. The goal there is to store the location alongside the test, in a way that Intellij is able discover without any call to weaver, by statically detecting an implicit conversion from "string" to "TestName". This is hacky, but works reasonably well. This allows for users to run individual tests. 
* The plugin uses coursier to dynamically resolve/fetch a CLI that can run weaver suites depending on the version of weaver referenced by the user's build. This prevents the user from having to have the plugin version match the weaver version in his project. 
* the Jetbrains marketplace is annoying in the sense that they're reviewing every single version, therefore I'm electing to publish the plugin to github, and a file to gh-pages that Intellij should be able to reference (and ping periodically to check for updates) 
* I'm commenting out the unit-tests of the plugin for now. @hwielenberg has got a lead on this : it appears that the test module ain't associated to a JRE. We'll figure it out later. 
* Documentation will come after we check the whole thing in an yet-to-be RC, so that we can take screenshots and stuff.


build.sbt
```diff
@@ -200,42 +224,69 @@ lazy val monix = crossProject(JSPlatform, JVMPlatform)
   )
 
 lazy val monixJVM = monix.jvm
-lazy val monixJS = monix.js
+lazy val monixJS  = monix.js
 
-lazy val cli = crossProject(JVMPlatform)
```
Baccata, line 205: I had wrongfully analysed the problem. We need a specific IntellijRunner that dumps the test data in a teamCity specific format.

modules/core/src/weaver/TestName.scala
```diff
@@ -0,0 +1,20 @@
...
+ *
+ * The implicit conversion from String is used as a mean
+ * for IDEs to detect the location of individual tests.
+ *
+ * The IDE is then able to request running a test at a specific
+ * location, using a CLI call. The framework will be able to lookup
+ * the test using the location (which should match with what the IDE
+ * knows)
+ */
+case class TestName(name: String, location: SourceLocation)
```
Baccata, line 14: @keynmol This is the main change in the PR. This allows to capture the location of tests, so that they can be looked-up by line number (which Intellij discovers by looking for the implicit conversions from "String" to "TestName").

modules/intellij/updatePlugins_template.xml
```diff
@@ -0,0 +1,18 @@
+<?xml version="1.0" encoding="UTF-8"?>
```
Baccata, line 1: This file is populated during publish stage and stored in a location that gets checked-in when publishing gh-pages.

modules/zio/src/weaver/ziocompat/suites.scala
```diff
@@ -22,18 +22,18 @@ abstract class BaseMutableZIOSuite[Res <: Has[_]](implicit tag: Tag[Res])
 
   private[this] type Test = ZIO[Env[Res], Nothing, TestOutcome]
 
-  protected def registerTest(name: String)(test: Test): Unit =
+  protected def registerTest(name: TestName)(test: Test): Unit =
```
Baccata, line 25: breaking change. The next release will be 0.5.0

scripts/releasePlugin.sc
```diff
@@ -0,0 +1,63 @@
+import $ivy.`com.lihaoyi::requests:0.6.5`
+import $ivy.`com.lihaoyi::upickle:0.9.5`
+
+@main
+def release(
```
Baccata, line 5: This script contains logic to publish the zipped plugin to github under the release.

modules/intellij/src/test/scala/weaver/intellij/testsupport/WeaverTestMethodTest.scala
```diff
@@ -0,0 +1,63 @@
...
+import com.intellij.codeInsight.daemon.impl.CollectHighlightsUtil
+import com.intellij.openapi.editor.Document
+import com.intellij.openapi.module.ModuleManager
+import com.intellij.openapi.util.Disposer
+import com.intellij.psi.PsiElement
+import org.jetbrains.plugins.scala.DependencyManagerBase.DependencyDescription
+// import org.junit.runner.RunWith
+// import org.junit.runners.AllTests
+
+// @RunWith(classOf[AllTests])
```
Baccata, line 14: @hwielenberg I'm commenting the tests for now

modules/intellij/src/main/scala/weaver/intellij/testsupport/WeaverTestConfigurationProducer.scala
```diff
@@ -0,0 +1,66 @@
...
+    WeaverTestConfigurationType.configurationFactory
+
+  override protected def suitePaths: Seq[String] =
+    WeaverTestFramework.baseSuitePaths
+
+  override protected def configurationName(
+      contextInfo: AbstractTestConfigurationProducer.CreateFromContextInfo): String =
+    contextInfo match {
+      case CreateFromContextInfo.AllInPackage(_, packageName) =>
+        s"ScalaTests in ''$packageName''"
```
keynmol, line 27: huehuehue

Baccata:
I've merged this into a 0.5.0-series branch. I want to try the publishing logic whilst deferring the merge for potentially more testing

PR #75, opened 2020-09-11
Colorise sample reports

keynmol:
**Published: https://weaver-site.surge.sh/weaver-test/docs/multiple_suites_failures**

Before             |  After
:-------------------------:|:-------------------------:
image |   image

### DIFF
--- a/modules/docs/src/main/scala/weaver/Output.scala
+++ b/modules/docs/src/main/scala/weaver/Output.scala
@@ -8,21 +8,21 @@ import scala.concurrent.ExecutionContext.Implicits.global
 
 object Output {
   implicit val cs          = IO.contextShift(global)
-  private val ansiColorRgx = "\u001b\\[([;\\d]*)m".r
 
-  def removeASCIIColors(str: String) =
-    ansiColorRgx.replaceAllIn(str, "")
-
-  def format(s: String) =
-    removeTrailingNewLine(removeASCIIColors(s.replace("repl.SessionApp", "")))
+  def format(s: String) = { 
+    Ansi2Html(removeTrailingNewLine(
+      removeTrailingNewLine(
+        s.replace("repl.MdocSessionApp", "").replace("(none:", "(MyTests.scala:")))
+      )
+  }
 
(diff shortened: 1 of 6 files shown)


modules/docs/src/main/scala/weaver/Output.scala
```diff
@@ -35,3 +35,79 @@ object Output {
     } yield value.reduceLeft(_ + "\n" + _)
   }
 }
+
+
+import fansi.Str
+
+object Ansi2Html extends Function1[String, String] {
```
keynmol, line 42: This is not a bulletproof implementation - most of it was just to make weaver's examples work.

It probably misses a ton of corner cases.

I can fix it if it ever becomes a problem - ideally I'd want to extract it as a separate mini-project.

modules/docs/src/main/scala/weaver/Output.scala
```diff
@@ -8,21 +8,21 @@ import scala.concurrent.ExecutionContext.Implicits.global
...
 
-  def removeASCIIColors(str: String) =
-    ansiColorRgx.replaceAllIn(str, "")
-
-  def format(s: String) =
-    removeTrailingNewLine(removeASCIIColors(s.replace("repl.SessionApp", "")))
+  def format(s: String) = { 
+    Ansi2Html(removeTrailingNewLine(
+      removeTrailingNewLine(
+        s.replace("repl.MdocSessionApp", "").replace("(none:", "(MyTests.scala:")))
```
keynmol, line 15: Fixes the names of suites and the filename `none`

keynmol:
Question: do we want to make more examples contain output rendered as such?

Baccata:
> Question: do we want to make more examples contain output rendered as such?

what would change if the answer was "yes" ?

keynmol:
> What would change if the answer was "yes" ?

Just more outputs, really. For things like https://disneystreaming.github.io/weaver-test/docs/expectations (with assertions) it might be good.

Not a big deal, but we have this ability now so can use it everywhere :D

Baccata (approved):

PR #76, opened 2020-09-12
WIP: GitHub Actions

bfdes:
Current strategy is to have three workflow files (test, website publish and artifact release), so that they may react to events differently. This means the Coursier and SBT cache isn't shared, and there is a little duplication of code. 

It may be possible to consolidate all the files into a single workflow [using `if` workflow syntax](https://docs.github.com/en/actions/reference/workflow-syntax-for-github-actions#jobsjob_idif). 

I've verified that the test workflow works on my fork, https://github.com/bfdes/weaver-test/pull/1, but the other two require secrets to be present. This might still be possible on my end.

### DIFF
--- a/.github/workflows/publish.yml
+++ b/.github/workflows/publish.yml
@@ -1,13 +1,12 @@
-name: Build and Test
+name: Publish documentation
 
 on:
   push:
-    branches:
-      - master
-  pull_request:
+    tags:
+      - v*
 
 jobs:
-  test:
+  publish:
     runs-on: ubuntu-latest
     steps:
     - name: Checkout repo
@@ -18,8 +17,11 @@ jobs:
       uses: actions/setup-java@v1
       with:
         java-version: 1.8
-    - name: Run tests
-      run: sbt ci
+    - name: Set up Node.js and Yarn
+      uses: actions/setup-node@v2-beta
+      with:
+        node-version: 12
+    - name: Publish docs
+      run: sbt docs/docusaurusPublishGhpages
       env:
-        JAVA_OPTS: "-XX:MaxMetaspaceSize=1g -Xms1g -Xmx2g -Xss2M"
-        SBT_OPTS: "-XX:+CMSClassUnloadingEnabled -XX:MaxMetaspaceSize=2g -Xmx2g"
+        GITHUB_DEPLOY_KEY: ${{ secrets.GITHUB_TOKEN }}
--- a/.github/workflows/publish.yml
+++ b/.github/workflows/publish.yml
@@ -2,6 +2,7 @@ name: Publish documentation
 
 on:
   push:
+    pull_request:
     tags:
       - v*
 
--- a/.github/workflows/publish.yml
+++ b/.github/workflows/publish.yml
@@ -4,7 +4,7 @@ on:
   push:
     tags:
       - v*
-  pull_request
+  pull_request:
 
 jobs:
   publish:
--- a/.github/workflows/publish.yml
+++ b/.github/workflows/publish.yml
@@ -2,9 +2,9 @@ name: Publish documentation
 
 on:
   push:
-    pull_request:
     tags:
       - v*
+  pull_request
 
 jobs:
   publish:
--- a/.github/workflows/release.yml
+++ b/.github/workflows/release.yml
@@ -9,7 +9,7 @@ on:
 
 jobs:
   publish:
-    run-on: ubuntu-latest
+    runs-on: ubuntu-latest
     steps:
     - name: Checkout repo
       uses: actions/checkout@v2
--- a/.github/workflows/release.yml
+++ b/.github/workflows/release.yml
@@ -1,4 +1,4 @@
-name: Publish documentation
+name: Release artifact
 
 on:
   push:
@@ -9,7 +9,7 @@ on:
 
 jobs:
   publish:
-    runs-on: ubuntu-latest
+    run-on: ubuntu-latest
     steps:
     - name: Checkout repo
       uses: actions/checkout@v2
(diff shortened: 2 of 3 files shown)


Baccata:
@bfdes thanks for starting to have a look into it. However we're gonna defer this work until the release of 0.5.0, as the build will be slightly more complex, and the priority is the very much the release over the change of CI tool. 

You can continue making progress on your fork by making sure the publishing of the website works, and you could publish to another artifact repository (bintray for instance) after changing the publishing org, just for a sanity check

.github/workflows/publish.yml
```diff
@@ -0,0 +1,29 @@
+name: Publish documentation
+
+on:
+  push:
+    branches:
+      - master
```
keynmol, line 6: I think we shouldn't release the site on master, as the documentation depends on current code and we don't want it to reflect unreleased version.

That said, docusaurus does support versioned documentation but I don't think the plugin supports that

.github/workflows/publish.yml
```diff
@@ -0,0 +1,29 @@
...
+      uses: coursier/cache-action@v3
+    - name: Set up JDK 1.8
+      uses: actions/setup-java@v1
+      with:
+        java-version: 1.8
+    - name: Set up Node.js and Yarn
+      uses: actions/setup-node@v2-beta
+      with:
+        node-version: 12
+    - name: Publish docs
```
keynmol, line 26: @Baccata that's what I was referring to, I've had success in publishing the site using a separate action and the github token that Actions creates, i.e.

```yaml
      - name: Build site
        run: sbt "buildSite --destination $(pwd)/_site"
      - name: Publish gh-pages
        uses: peaceiris/actions-gh-pages@v3
        with:
          github_token: ${{ secrets.GITHUB_TOKEN }}
          publish_dir: ./_site
```

(ignore the actual sbt command, that's from my project)

We might do it this way to avoid having a deploy key in the secrets.

.github/workflows/release.yml
```diff
@@ -0,0 +1,26 @@
+name: Release artifact
+
+on:
+  push:
+    branches:
+      - master
```
keynmol, line 6: Would be good to verify that dynver works correctly and publishes a snapshot version to Snapshots Sonatype.

In my projects I use https://github.com/olafurpg/sbt-ci-release which is designed for this flow, but I don't want to mess with this part of the build just yet

.github/workflows/test.yml
```diff
@@ -0,0 +1,25 @@
...
+    runs-on: ubuntu-latest
+    steps:
+    - name: Checkout repo
+      uses: actions/checkout@v2
+    - name: Restore Coursier cache
+      uses: coursier/cache-action@v3
+    - name: Set up JDK 1.8
+      uses: actions/setup-java@v1
+      with:
+        java-version: 1.8
```
keynmol, line 20: I wonder if we should make it into a build matrix for JDK 8 and 11, as I've seen some funky stuff of library updates sneaking through that don't work on older JVMs

.github/workflows/release.yml
```diff
@@ -0,0 +1,26 @@
...
+    - name: Checkout repo
+      uses: actions/checkout@v2
+    - name: Restore Coursier cache
+      uses: coursier/cache-action@v3
+    - name: Set up JDK 1.8
+      uses: actions/setup-java@v1
+      with:
+        java-version: 1.8
+    - name: Release artifact
+      run: sbt release
```
keynmol, line 23: What confuses me is I don't understand where sbt comes from :-/ 

I usually use https://github.com/olafurpg/setup-scala which sets it up, but not sure which action brings it in here.

keynmol:
@bfdes sorry for the delay on this one.

I'm happy for us to unfreeze this work, test it with latest master and we can discuss with @Baccata my comments on the PR.

.github/workflows/release.yml
bfdes: IIRC the action to setup Java included it.

keynmol: Hah. Crazy. I think it's actually available on Linux by default: https://github.com/actions/virtual-environments/blob/51d02f26f42e53bb963bb6bc9754501b28a3d2ba/images/linux/scripts/installers/sbt.sh

bfdes: Hmm its in the Windows image installer scripts too, but not sure about the Mac one. 
We should have a matrix build to test on every platform as well as supported JDKs, right?

keynmol: That'd be overkill, tbh, builds might get crazy long - plus most of the stuff that can go wrong is in SBT's and effect system's realm, and they probably do have rich mbuild matrices.

JDKs on the other hand I don't think people are as vigilant.

bfdes: Build time should not be an issue as the builds run in parallel.

Having said that, I found that the macOS image does not come with SBT so I'll simplify it and just run on Ubuntu.

.github/workflows/publish.yml
```diff
@@ -0,0 +1,29 @@
...
+      run: sbt docs/docusaurusPublishGhpages
```
keynmol, line 27:
```suggestion
      run: amm scripts/releaseSite.sc
```

https://github.com/disneystreaming/weaver-test/blob/master/.drone.yml#L49

Also means we need to install ammonite here.

.github/workflows/test.yml
```diff
@@ -0,0 +1,22 @@
...
+      uses: coursier/cache-action@v3
+    - name: Set up JDK
+      uses: actions/setup-java@v1
+      with:
+        java-version: ${{matrix.javaVersion}}
+    - name: Run tests
+      run: sbt ci
+      env:
+        JAVA_OPTS: "-XX:MaxMetaspaceSize=1g -Xms1g -Xmx2g -Xss2M"
+        SBT_OPTS: "-XX:+CMSClassUnloadingEnabled -XX:MaxMetaspaceSize=2g -Xmx2g"
```
keynmol, line 22: We're now up to 7G on the heap to run the tests :D

https://github.com/disneystreaming/weaver-test/blob/master/.drone.yml#L14

keynmol:
As our matrix has grown considerably, I had to change the approach significantly - and will continue in #230 

Thanks Bruno and sorry it took so long to get back to this!

Issue #81, opened 2020-09-16
Investigate integration with monix-bio


keynmol:
[monix-bio](https://github.com/monix/monix-bio) is in 1.x version, so we can probably assume that it's stable.

It's worth looking into integrating with it to provide full coverage of IO monads in existence. Worth checking with maintainers if the API is considered stable.

Baccata:
Already pinged @bfdes as he's done the monix Task integration. Most of it will be copy/paste.

Baccata:
Addressed in #83

PR #82, opened 2020-09-18
Relative paths during macro capture

Baccata:
Browser-runtimes are not filesystem aware. This change should allow
folks to use weaver for dom testing.

PR #83, opened 2020-09-18
Monix BIO support

bfdes:
Attempt to address https://github.com/disneystreaming/weaver-test/issues/81.

This PR introduces: 
- the compatibility module 
- documentation (on a separate page to that of Monix compatibility)
- refactoring of the way we perform assertions in all compatibility module tests


build.sbt
```diff
@@ -203,6 +207,22 @@ lazy val monix = crossProject(JSPlatform, JVMPlatform)
...
 
+lazy val monixBio = crossProject(JSPlatform, JVMPlatform)
+  .crossType(CrossType.Pure)
+  .in(file("modules/monixBio"))
+  .dependsOn(core, framework % "test->compile")
+  .configure(WeaverPlugin.profile)
+  .settings(WeaverPlugin.simpleLayout)
+  .settings(
+    libraryDependencies ++= Seq(
+      "io.monix" %% "monix-bio" % "1.0.0"
```
Baccata, line 218: There's a mistake here which causes the build to fail. I'll let you find what it is 😄

build.sbt
```diff
@@ -203,6 +207,22 @@ lazy val monix = crossProject(JSPlatform, JVMPlatform)
...
+    ),
+    scalaJSLinkerConfig ~= { _.withModuleKind(ModuleKind.CommonJSModule) }
```
keynmol, line 220: I'm hoping we can merge #82 before this, so that we don't have to emit commonjs modules exclusively anymore.

build.sbt
bfdes: 🤦 I'm guessing I need `%%%` to get the Scala.js source, too.

bfdes:
Looks like I've caused the tests to hang.

Baccata:
@bfdes indeed. On the bright side, it seems to be reproducible locally (`monixBioJS/test` in sbt)

modules/monixBio/test/src/weaver/monixbiocompat/IOSuiteTest.scala
```diff
@@ -0,0 +1,58 @@
...
+    test("example test") {
+      Task.raiseError(new RuntimeException("oh no"))
+    }
+  }
+
+  object TestWithExceptionInExpectation extends SimpleIOSuite {
+    test("example test") {
+      for {
+        _ <- Task.unit
+      } yield throw new RuntimeException("oh no")
```
Baccata, line 41: @avasil, 👋 , pinging your for advice. We're adding support for monix-bio, and this throw seems to make it hang forever. Is it to be expected, or is it a bug ? 

Now I'm aware this is very much the fault of the user, but being a maintainer of a test framework, and knowing how liberal people are with tests in general, I know many will likely mess up and come to us for explanation 😅 

Is there anything on our side we could do to protect against it ?

Baccata: Apologies, wrong analysis, the problem seem to be lying elsewhere

bfdes:
Also occurs when running `monixBioJVM/test`, and disappears when `TestWithExceptionInExpectation` is commented out.

modules/monixBio/test/src/weaver/monixbiocompat/IOSuiteTest.scala
```diff
@@ -0,0 +1,58 @@
...
+  List(
+    TestWithExceptionInTest,
+    TestWithExceptionInExpectation,
+    TestWithExceptionInInitialisation
+  ).foreach { testSuite =>
+    test(s"fail properly in ${testSuite.getClass.getSimpleName}") {
+      for {
+        (_, events) <- DogFood.runSuite(testSuite).to[Task]
+      } yield {
+        val event = events.headOption.get
```
Baccata, line 18: Right, so my analysis seems to actually be correct : monix-bio really doesn't like throws in `map`, and we happened to potentially throw in this map which threw me off. @bfdes, can you change the content of this yield to the following : 

```scala
        val maybeEvent = events.headOption
        val maybeThrowable = maybeEvent.flatMap { event =>
          if (event.throwable().isDefined()) Some(event.throwable().get)
          else None
        }
        val maybeStatus = maybeEvent.map(_.status())
        expect(maybeStatus.contains(Status.Error)) &&
        expect(maybeThrowable.map(_.getMessage).contains("oh no"))
```

modules/monixBio/test/src/weaver/monixbiocompat/IOSuiteTest.scala
Baccata: @avasil, actually I think my first analysis was correct : monix-bio seems to get stuck when throws are performed in a `map` call, even when the error channel expect throwable. 

Any guidance on how to recover would be appreciated

Avasil: Yes, if there is an error in `map`, it is treated as an unexpected error (in a different channel). 
It can catch off guard when using `IO[Throwable, A]`

I assume that it hangs because the error is not handled properly - you can use `redeemCause` to get access to both types of errors. If you are able to override `test` method, you could do `redeemCauseWith(c => IO.raiseError(c.toThrowable), x => IO.now(x))`

I think I should add an operator that changes all terminal errors to typed errors if the type is `Throwable`, like ZIO's absorb.

modules/monixBio/test/src/weaver/monixbiocompat/IOSuiteTest.scala
bfdes: I ported these assertions from Monix compatibility tests (which were in turn ported from ZIO compatibility tests). Changing the content of the yield does not seem to fix the issue, which as @Avasil says below seems to be down to the throw in `TestWithExceptionInExpectation` -- is this what you expected?

Going to do it anyhow for the sake of cleaner code (and I'll also amend the other two suites).

modules/monixBio/test/src/weaver/monixbiocompat/IOSuiteTest.scala
bfdes: @Baccata I provided an implementation of `Test.apply` in the Monix BIO compat module with the additional error checks @Avasil mentioned and this seemed to resolve the issue. Commit incoming. 

It required minimal changes in core code, but please tell me if there is a better way to go about it.

modules/monixBio/src/weaver/monixbiocompat/suites.scala
```diff
@@ -13,10 +15,49 @@ trait BaseIOSuite { self: ConcurrentEffectSuite[Task] =>
...
+  def apply(
+      name: String,
+      f: Log[Task] => Task[Expectations]): Task[TestOutcome] =
+    for {
+      ref   <- Ref[Task].of(Chain.empty[Log.Entry])
+      start <- IO.clock.realTime(MILLISECONDS)
+      res <- IO
+        .defer(f(Log.collected[Task, Chain](ref)))
+        .map(Result.fromAssertion)
+        .redeemCauseWith(c => IO.raiseError(c.toThrowable), IO.now)
```
Avasil, line 36: Looks like it could be `redeemCause(c => Result.from(c.toThrowable), identity)` without `handleError` below

bfdes: Thanks, addressed in d2a587c05c8e3af3512ab6bd744fdf64ec0bb396.

modules/monixBio/test/src/weaver/monixbiocompat/IOSuiteTest.scala
Baccata: > I ported these assertions from Monix compatibility tests 

yup, you're definitely not at fault there 😄 . Ironically, this code was potentially throwing, which added to the complexity of the analysis since the problem appeared not only in this test, but also in the test this test was testing (yes, it's confusing). 

> Changing the content of the yield does not seem to fix the issue, which as @Avasil says below seems to be down to the throw in TestWithExceptionInExpectation -- is this what you expected? 

Yeah I'm not expecting this change to resolve the issue, but rather make it slightly easier to pinpoint in the future if it was to re-occur in other implementations.

modules/core/src/weaver/suites.scala
```diff
@@ -91,12 +91,13 @@ trait MutableFSuite[F[_]] extends ConcurrentEffectSuite[F]  {
   def pureTest(name: String)(run : => Expectations) :  Unit = registerTest(name)(_ => Test(name, effect.delay(run)))
   def simpleTest(name:  String)(run: => F[Expectations]) : Unit = registerTest(name)(_ => Test(name, effect.suspend(run)))
   def loggedTest(name: String)(run: Log[F] => F[Expectations]) : Unit = registerTest(name)(_ => Test(name, log => run(log)))
-  def test(name: String) : PartiallyAppliedTest = new PartiallyAppliedTest(name)
+  def test(name: String) : PartiallyAppliedTest = new PartiallyAppliedTest { val testName: String = name }
```
Baccata, line 94: I don't think the change is absolutely necessary here, in the monix one you can do 

``` 
def test(name: String) : PartiallyAppliedTest = new this.PartiallyAppliedTest(name) 
class PartiallyAppliedTest(name) extends super.PartiallyAppliedTest(name) 
```

modules/monixBio/test/src/weaver/monixbiocompat/IOSuiteTest.scala
bfdes: > Ironically, this code was potentially throwing, which added to the complexity of the analysis since the problem appeared not only in this test, but also in the test this test was testing (yes, it's confusing).

😆 

Refactored in fb0ad99b3389f7411a3fcf0a1ec61446cf71ff84 I think.

modules/core/src/weaver/suites.scala
bfdes: Addressed in c82f37f572c601d2209501fded780c7abee49130. I had to rename the subclass because of a compiler warning that this kind of shadowing is deprecated.

docs/monix_bio_usage.md
```diff
@@ -0,0 +1,93 @@
+---
+id: monix_bio
+title: Monix BIO usage
+---
+
+Testing Monix BIO programs is very similar to testing regular Monix progams.
```
bfdes, line 6: @Avasil do you think this introduction makes sense? I'm guessing most people using the module will know how `Task` relates to `IO`, but I thought it was worth reiterating.

modules/monixBio/src/weaver/monixbiocompat/package.scala
```diff
@@ -0,0 +1,6 @@
+package weaver
+
+package object monixbiocompat {
+  type IOSuite       = MutableIOSuite
```
bfdes, line 4: Suites are named after the effect type, `IO`, rather than the alias `Task`.

This is consistent with how the ZIO suites were named.

bfdes:
I updated the README as well as the website. Not sure if the latter is strictly necessary, given the similarity to the API of the Monix module.

modules/monixBio/src/weaver/monixbiocompat/package.scala
Baccata: cats-effect is moving towards making its typeclasses parametric in the Error type. It doesn't mean there will be a bi-functor hierarchy, however we will be able to let users parameterise the suites : `IOSuite` will become `IOSuite[E]` for instance, which will make the naming less surprising.

Baccata:
>  Not sure if the latter is strictly necessary, given the similarity to the API of the Monix module.

I'm happy for both to be present as some people go directly to the website and wouldn't necessarily notice the README changes. Plus it puts monix-bio on an equal footing with the rest of the effect types, which I think is desirable

Baccata (approved):
Thank you !

Avasil:
Thank you and I'm glad to see Monix BIO doc section too, it will be more discoverable :)

PR #96, opened 2020-10-03
Expecty 0.14.1 / expectations rework

Baccata:
Update expecty to 0.14.0
    
* Source locations for "expect" are now sourced directly from the macro
* As a result, "SingleExpectation" is not needed anymore
* A new `expect.all` macro method allows to reduce boilerplate
* AssertionException now contains a (non-empty) list of locations.
* Expectations now has a "trace" method to append a location to all its errors.
* Addition of an "here" method that sources a location without being impacted by the implicit scope.

### DIFF
--- a/docs/expectations.md
+++ b/docs/expectations.md
@@ -18,6 +18,7 @@ object MySuite2 extends SimpleIOSuite {
   }
 
   pureTest("Varargs composition") {
+    // expect(1 + 1 == 2) && expect (2 + 2 == 4) && expect(4 * 2 == 8)
     expect.all(1 + 1 == 2, 2 + 2 == 4, 4 * 2 == 8)
   }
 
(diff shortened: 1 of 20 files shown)


Baccata:
This will fix the build : https://github.com/eed3si9n/expecty/pull/31, just need to wait for a release

docs/expectations.md
```diff
@@ -17,6 +17,10 @@ object MySuite2 extends SimpleIOSuite {
     expect(1 != 2) and expect(2 != 1) or expect(2 != 3)
   }
 
+  pureTest("Varargs composition") {
+    expect.all(1 + 1 == 2, 2 + 2 == 4, 4 * 2 == 8)
```
keynmol, line 21: I think multiline here would read better

keynmol (approved):

PR #110, opened 2020-10-19
[WIP] Start scalacheck-effect integration

### LINKED ISSUE #79: Integrate with scalacheck-effect
https://github.com/typelevel/scalacheck-effect

This has existed for only a month but respects Scalacheck idioms much better than our hand-rolled  integration that didn't account for the "Prop" since they were creating a synchronous boundary.

keynmol:
Closes #79 

This is just a WIP.

I want to discuss what sort of UX we present to the user of those traits, how we can make the rendering of failures better (tbh I actually like a combination of expecty's output and argument output)

image

My issue here is mostly the source location being super wrong.

And also the amount of configuration for Scalacheck we expose on a per-test/per-spec basis.

### DIFF
--- a/build.sbt
+++ b/build.sbt
@@ -172,7 +172,8 @@ lazy val scalacheck = crossProject(JSPlatform, JVMPlatform)
   .settings(WeaverPlugin.simpleLayout)
   .settings(
     libraryDependencies ++= Seq(
-      "org.scalacheck" %%% "scalacheck" % "1.14.3"
+      "org.scalacheck" %%% "scalacheck"        % "1.14.3",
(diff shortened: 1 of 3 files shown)


Baccata:
For the sake of the traceability : I had started a similar integration previously, and had stopped when after realising that the way scalacheck-effect is written applies effects sequentially. Weaver's integration uses fs2 to apply them in parallel (depending on tweakable settings). 

Considering weaver's whole premise is to parallelise as much as possible, I felt like integrating this was maybe not a good idea. I haven't made my mind up yet.

keynmol:
We can revisit if anyone explicitly asks for it, but overall I agree with you.

PR #143, opened 2020-12-06
Framework revamp

Baccata:
This work aims at narrowing the gap required to be jumped for CE3 compatibility. 

CE3 is removing the abstraction allowing to "unsafeRun" effects. It is also removing the abstraction allowing to switch from one effect to cats'IO. These features are currently heavily used by weaver, which currently runs one IO per suite. 

We're making the decision to drastically change the internals of the framework in order to support an upgrade to CE3. Overall, this means that weaver now comprises 4 different test frameworks, one per effect-type currently existing in the ecosystem (cats-effect, monix, monix-bio, zio). All of them delegate to a generic one, which runs a *single IO* per run (ie compile-unit), and communicates with the build tool via an ad-hoc promise/concurrent queues based protocol. However each framework has a thin, effect-specific layer associated to it that dictates how things run (essentially calling the `unsafeRun` equivalent of the effect type). 

This means that weaver is becoming more respectful of the choices of its users, as cats-IO will not be run, at all, if users elect to use monix. It also means that features such as "Global Resources" can now be interacted with in terms of whichever effect type is preferred by the user.


keynmol:
image

LGTM, let's ship :D

project/WeaverPlugin.scala
```diff
@@ -102,8 +102,8 @@ object WeaverPlugin extends AutoPlugin {
     "-Ywarn-unused:locals",          // Warn if a local definition is unused.
     "-Ywarn-unused:patvars",         // Warn if a variable bound in a pattern is unused.
     "-Ywarn-unused:privates",        // Warn if a private member is unused.
-    "-Ywarn-value-discard",          // Warn when non-Unit expression results are unused.
-    "-Xfatal-warnings"               // Fail the compilation if there are any warnings.
+    "-Ywarn-value-discard"           // Warn when non-Unit expression results are unused.
+    // "-Xfatal-warnings"               // Fail the compilation if there are any warnings.
```
keynmol, line 106: ah. Should re-enable this and do some clean up.

PR #158, opened 2020-12-17
Scala 3.0.0-M3

keynmol:
~Super early version, most compilation works (apart from scalacheck)~ scalacheck works

~Dirty comments left here'n'there~

Lots of compat layers, ~lots of `withDottyCompat`~

~Given it's a test framework, it might not be a horrible idea to publish versions with withDottyCompat.~

~Tested with locally published https://github.com/eed3si9n/expecty/pull/35~ 0.15.0 is published

TODO:

- [x] Improve the Build Matrix in docs


modules/core/src/weaver/Runner.scala
```diff
@@ -37,15 +37,15 @@ class Runner[F[_]: CECompat.Effect](
...
 
-  def complete(channel: Channel[SpecEvent]): F[Unit] =
-    channel.enqueue1(None) // We are done !
+  private def complete(channel: Channel[SpecEvent]): F[Unit] =
+    channel.enqueue(None) // We are done !
 
   // Recursively consumes from a channel until a "None" gets produced,
   // indicating the end of the stream.
-  def consume(
+  private def consume(
```
keynmol, line 48: should probably make this package private, I know I use it internally.

modules/scalacheck/test/src/weaver/scalacheck/PropertyDogFoodTest.scala
```diff
@@ -31,7 +34,7 @@ object PropertyDogFoodTest extends IOSuite {
   // 100 checks sleeping 1 second each should not take 100 seconds
   test("Checks are parallelised") { dogfood =>
     for {
-      (_, events) <- dogfood.runSuite(Meta.ParallelChecks)
+      events <- dogfood.runSuite(Meta.ParallelChecks).map(_._2)
```
keynmol, line 37: IIRC that's because better-monadic-for is not available on Scala 3.

Baccata: That makes me SO GODDAMN ANGRY

build.sbt
```diff
@@ -7,16 +7,17 @@ ThisBuild / commands += Command.command("ci") { state =>
...
     "core/publishLocal" :: state
 }
 
 ThisBuild / commands += Command.command("fix") { state =>
-  "scalafmtAll" ::
-    "scalafmtSbt" ::
-    "scalafix" ::
-    "test:scalafix" :: state
+  "scalafix" ::
+    "test:scalafix" ::
```
keynmol, line 18: because scalafix can actually change the formatting :-/

build.sbt
```diff
@@ -75,17 +75,19 @@ lazy val core = projectMatrix
...
-      "org.portable-scala"   %%% "portable-scala-reflect" % "1.0.0"
+      "com.eed3si9n.expecty" %%% "expecty"                % "0.15.0",
+      ("org.portable-scala"  %%% "portable-scala-reflect" % "1.0.0").withDottyCompat(
+        scalaVersion.value)
     ),
     libraryDependencies ++= {
       if (virtualAxes.value.contains(VirtualAxis.jvm))
         Seq(
-          "org.scala-js" %%% "scalajs-stubs" % "1.0.0" % "provided"
+          ("org.scala-js" %%% "scalajs-stubs" % "1.0.0" % "provided").withDottyCompat(
```
keynmol, line 85: this dep is like 2 annotations, may be we can copypaste it?

build.sbt
```diff
@@ -217,7 +221,8 @@ lazy val specs2 = projectMatrix
     name := "specs2",
     testFrameworks := Seq(new TestFramework("weaver.framework.CatsEffect")),
     libraryDependencies ++= Seq(
-      "org.specs2" %%% "specs2-matcher" % "4.10.5"
+      ("org.specs2" %%% "specs2-matcher" % "4.10.5").withDottyCompat(
```
keynmol, line 224: I'm not seeing a 3.0.0-M3 release yet, so this might have to become onlyScala2 for the time being.

modules/core/src/weaver/TestErrorFormatter.scala
```diff
@@ -64,6 +64,8 @@ object TestErrorFormatter {
             append(Snip(excl))
           case (Some(_: Element), Some(excl)) =>
             append(Snip(excl))
+
+          case (Some(Snip(_)), Some(_)) => throw new Exception("wat")
```
keynmol, line 68: temporary, I will fix this pattern match.

project/WeaverPlugin.scala
```diff
@@ -130,28 +175,44 @@ object WeaverPlugin extends AutoPlugin {
...
-    ThisBuild / scalacOptions ++= Seq(
-      // Note, this is used by the doc-source-url feature to determine the
-      // relative path of a given source file. If it's not a prefix of a the
-      // absolute path of the source file, the absolute path of that file
-      // will be put into the FILE_SOURCE variable, which is
-      // definitely not what we want.
-      "-sourcepath",
-      file(".").getAbsolutePath.replaceAll("[.]$", "")
-    ),
+    // ThisBuild / scalacOptions ++= {
```
keynmol, line 178: need to fix this (it breaks on dotty)

project/WeaverPlugin.scala
```diff
@@ -28,47 +31,88 @@ object WeaverPlugin extends AutoPlugin {
...
-          configureScalaJSProject(_).settings(versionOverrideForCE3)
-        )
+    def add(
+        scalaVersions: Iterable[String],
+        platform: VirtualAxis.PlatformAxis,
+        catsEffectAxis: CatsEffectAxis): ConfigureX = {
+      scalaVersions.map(addOne(_, platform, catsEffectAxis)).reduce(_ andThen _)
+    }
+
+    def onlyCatsEffect2(
```
Baccata, line 83: let's try and make this a bit more conceptually consistent, by changing it to : 

```
def sparse(
  withCE3 : Boolean = true, 
  withJS: Boolean = true, 
  withScala3: Boolean = true
) 
```

modules/core/src/weaver/Expectations.scala
```diff
@@ -6,6 +6,24 @@ import cats.data.{ NonEmptyList, Validated, ValidatedNel }
 import cats.effect.Sync
 import cats.syntax.all._
 
+trait Expectable[F[_], A] {
+  def lift(a: A): F[Expectations]
+}
+
+object Expectable {
+  def apply[F[_], B](implicit ev: Expectable[F, B]): Expectable[F, B] = ev
+  implicit def syncExpectable[F[_]: Sync]: Expectable[F, Expectations] =
```
Baccata, line 15: this should be constrained to `Applicative`, with `pure` being used

keynmol: love me a bit of Sync here and there. Good point, not sure why I needed this.

modules/core/src/weaver/Expectations.scala
```diff
@@ -6,6 +6,24 @@ import cats.data.{ NonEmptyList, Validated, ValidatedNel }
 import cats.effect.Sync
 import cats.syntax.all._
 
+trait Expectable[F[_], A] {
```
Baccata, line 9: I'm not sure this should be in `core`, since we only use it for scalacheck. Maybe we should rename it to `Prop`

build.sbt
Baccata: I'm fine with that

build.sbt
Baccata: mmm not sure how we could do it without polluting the user's classpath tbh. I think having the dotty compat on this one is low risk for the time being, since it's a compile-only thing.

modules/core/src/weaver/Expectations.scala
keynmol: May be we can move it into `Checkers` itself?

build.sbt
keynmol: i.e. with making it scala 2 only?

modules/core/src/weaver/Expectations.scala
Baccata: in the companion object, sure

build.sbt
Baccata: > i.e. with making it scala 2 only?

yes

Baccata (approved):

PR #164, opened 2020-12-20
Fix site release script (hopefully)

keynmol:
1. Call `main` on CI (should we remove `doInstallSSH` from main methods?)
2. Look at correct folder for mdoc (I regret making docs a project matrix :D but alternative would've been faffier)


Baccata:
> should we remove doInstallSSH from main methods? 

I do think so

Baccata (approved):

PR #172, opened 2020-12-23
Feature Request: zio now uses fiber ref for logging

lukecollier:
**Description**:

At my current team we're logging a lot in our integration tests, weaver has per test conditional logging built in to allow wiring weavers logging into these tests would be a highly desirable feature.

This PR uses `zio.FiberRef` over providing a layer to each run. This seems to be more idiomatic to ZIO and also gives us the potential for some very interesting functionallity,

For example this PR now exposes `LogModule` feeding it into the environment for the tests by some way of an adapter layer.
extended example:
```scala
import zio._
import weaver.ziocompat._
import weaver.{SourceLocation, Log}

object OurLogger {
  trait Service {
    def log(msg: String, level: String): UIO[Unit]
  }
  def log(msg: String, level: String): URIO[Has[Service], Unit] = ZIO.accessM[Has[Service]](_.get.log(msg, level))
}


object OurApp {
  class Program(logger: OurLogger.Service) extends OurApp.Service {
    def run(): UIO[Unit] = logger.log("hello", "debug") *> logger.log(", world", "debug")
  }

  val program: URLayer[Has[OurLogger.Service], Has[OurApp.Service] =ZLayer.fromService[LogModule.Service, OurLogger.Service](new Program(_))
  }

  trait Service {
    def run(): UIO[Unit]
  }
  def run(): URIO[Has[Service], Unit] = 
    ZIO.accessM[Has[Service]](_.get.run())
}

object OurTest extends ZIOSuite[Has[OurApp.Service]] {

  val loggingAdapterLayer: ZLayer[LogModule, Has[OurLogger]] = ZLayer.fromService[LogModule.Service, OurLogger.Service]{
    weaverLogger => new OurLogger {
      def log(msg: String, level: String): UIO[Unit] = 
        weaverLogger.log(Log.Entry(0L, msg, Map.empty, Log.Level(level), None, Log.SourceLocation("", "", 0))
    }
  }
  val sharedLayer: ZLayer[ZEnv with LogModule, Throwable, OurApp.Service] = loggingAdapterLayer >>> OurApp.program

  test("can run and log") {
    for {
      _ <- OurApp.run()
      logs <- LogModule.logs
    } yield expect(logs.size == 2)
  }
}

```


Baccata:
In principle, I don't disagree with this change, but a few tests would have to be added in order for me to approve this. 

In particular, you'd have to run a suite manually (by calling [spec](https://github.com/disneystreaming/weaver-test/blob/master/modules/core/zio/src/weaver/ziocompat/suites.scala#L36)) to make sure that logs would indeed belong to each particular test, especially when "maxParallelism" is set to 1 on the suite (which fs2 might be optimising against by avoiding to spawn new fibers, which would have the effect of all tests running on the same fiber and sharing the same state, defeating the purpose)

modules/core/zio/src/weaver/ziocompat/suites.scala
```diff
@@ -42,8 +45,15 @@ abstract class BaseMutableZIOSuite[Res <: Has[_]](implicit tag: Tag[Res])
       }
       if (filteredTests.isEmpty) Stream.empty // no need to allocate resources
       else {
-        val suiteLayer = sharedLayer.passthrough
         for {
+          ref <-
+            Stream.resource(
+              FiberRef.make(Chain.empty[Log.Entry]).toManaged_.toResourceZIO)
+          perTestLayer: RLayer[ZEnv, LogModule with ZEnv] =
```
Baccata, line 52: it ain't specific to a test anymore (you've renamed PerTestEnv to Env, you should rename this value too)

modules/core/zio/src/weaver/ziocompat/suites.scala
```diff
@@ -42,8 +45,15 @@ abstract class BaseMutableZIOSuite[Res <: Has[_]](implicit tag: Tag[Res])
       }
       if (filteredTests.isEmpty) Stream.empty // no need to allocate resources
       else {
-        val suiteLayer = sharedLayer.passthrough
         for {
+          ref <-
+            Stream.resource(
+              FiberRef.make(Chain.empty[Log.Entry]).toManaged_.toResourceZIO)
```
Baccata, line 51: This is overkill, the following should suffice : `Stream.eval(FiberRef.make(Chain.empty[Log.Entry])`

lukecollier:
> In particular, you'd have to run a suite manually (by calling spec) 

Can this not achieved by using the DogFoodz module in the zio tests?

Baccata:
> Can this not achieved by using the DogFoodz module in the zio tests?

It'd be harder, considering the Dogfoodz construct exercises the test framework itself as opposed to just the suite. Exercising the suite will make it easier to test that logs belong to separate tests, since you'll have access to a higher-level api

lukecollier:
Not had much time recently but will add the checks ASAP

modules/framework/zio/test/src/weaver/ziocompat/MutableZIOSuiteTest.scala
```diff
@@ -75,6 +75,33 @@ object ZIOSuiteTest extends ZIOSuite[KVStore with DogFoodz] {
     }
   }
 
+  test("logs contain only the logs from each test") {
+    FiberRefLogTest.spec(List.empty)
```
lukecollier, line 79: Is this the kind of test you were after @Baccata ?

Baccata: Yes, but I'd also like some tests showcasing the benefit of using a FiberRef based solution (you could re-use the example you have in the description of this PR)

modules/framework/zio/test/src/weaver/ziocompat/MutableZIOSuiteTest.scala
```diff
@@ -75,6 +76,99 @@ object ZIOSuiteTest extends ZIOSuite[KVStore with DogFoodz] {
...
+
+    val program: URLayer[Has[SomeLogger.Service], Has[SomeApp.Service]] =
+      ZLayer.fromService[SomeLogger.Service, SomeApp.Service](new Program(_))
+
+    trait Service {
+      def run(): Task[Unit]
+    }
+    def run(): RIO[Has[Service], Unit] =
+      ZIO.accessM[Has[Service]](_.get.run())
+  }
```
lukecollier, line 151: New test that shows a logger adapter in use.

lukecollier:
Is there possibility of getting this in an upcoming release? 

I can add a backwards compatability test but I believe ZIO handles that case for us (i.e `sharedLayer: [ZEnv with LogModule, Throwable, ?]` ~= `sharedLayer: [ZEnv, Throwable, ?]`

Very keen to use it on a project to reduce our log noise by 99%

Baccata:
@lukecollier the implications are significant enough that I need to give it some proper thoughts, as this would break feature parity with other modules. Therefore I'm gonna ask you for a little bit of patience as I find the time to collect my thoughts a little bit on this, and am rather busy this week and the next.

If it's that significant to your usecase that your absolutely need it by this week, you could copy the zio module locally in your project and use that, it will just work :)

PR #175, opened 2021-01-05
Prevent the user from having to ascribe type with Checkers

Baccata:


modules/scalacheck/src/weaver/scalacheck/Checkers.scala
```diff
@@ -120,9 +117,11 @@ trait Checkers[F[_]] {
       .takeRight(1) // getting the first error (which finishes the stream)
       .compile
       .last
-      .map {
-        case Some(status) => status.endResult
-        case None         => Expectations.Helpers.success
+      .map { (x: Option[Status[A]]) =>
+        x match {
```
Baccata, line 121: for some obscure reason, the compiler wasn't happy anymore 🤷‍♂️

RobinJDCox (approved):
Thanks!

keynmol (approved):

PR #176, opened 2021-01-08
Adds support for pretty printed equality comparison 

Baccata:
Leveraging Expecty's awesomeness one more time. 

```scala
expect.same("foo foo foo", "foo bar foo")
```

Screenshot 2021-01-10 at 16 28 49

Just need to wait for [this](https://github.com/eed3si9n/expecty/pull/41) to be released


keynmol:
hate to say it, but yo do you want to add a dogfood test for how this is rendered?

keynmol:
And honestly, anything stopping us from using this in dog food tests? instead of my bad `expectEquals` (although I don't know how this works with whitespace)

Baccata:
> hate to say it, but yo do you want to add a dogfood test for how this is rendered?

will do over the weekend, or next week 

> And honestly, anything stopping us from using this in dog food tests 

my own laziness ?

modules/core/src/weaver/Expectations.scala
```diff
@@ -134,15 +134,6 @@ object Expectations {
     def inEach[L[_], A](la: L[A])(f: A => Expectations)(
         implicit L: Foldable[L]): Expectations = forEach(la)(f)
 
-    def verify(condition: Boolean, hint: String)(
-        implicit pos: SourceLocation): Expectations =
-      if (condition) success
```
keynmol, line 139: sure you want to remove this?

Baccata: oops, probably not :s

keynmol (approved):

modules/core/src/weaver/internals/ExpectyListener.scala
```diff
@@ -1,11 +1,13 @@
 package weaver
+package internals
 
 import cats.data.{ NonEmptyList, ValidatedNel }
 import cats.syntax.all._
 
 import com.eed3si9n.expecty._
 
-class ExpectyListener extends RecorderListener[Boolean, Expectations] {
+private[weaver] class ExpectyListener
```
keynmol, line 9: I remember you asked me on the scala 3 PR if making it package private will work outside of weaver package (and it does, but we're not sure why)

Baccata: Are you not confusing the `ExpectyListener` class with the `Expecty` class ?

keynmol: image

PR #177, opened 2021-01-11
Adds support for memoised resources

Baccata:
This adds a construct allowing for lazy allocation/eager de-allocation of resources, which could come handy when users use global resources to keep computationally heavy constructs in the background (distributed log consumers, and so on). 

It is currently not cancel-safe, but I intend to specialise it to CE2/CE3 in a later unit of work

### DIFF
--- a/modules/core/src/weaver/MemoisedResource.scala
+++ b/modules/core/src/weaver/MemoisedResource.scala
@@ -29,7 +29,7 @@ private class MemoisedResource[F[_]: Concurrent, A] {
         compute <- ref.modify {
           case Uninitialised =>
             val newState = InUse(valuePromise, finaliserPromise.get.flatten, 1)
-            val compute = resource.allocated.attempt.flatMap {
+            val compute = Concurrent[F].attempt(resource.allocated).flatMap {
               case Right((a, fin)) => for {
                   _ <- valuePromise.complete(Right(a))
                   _ <- finaliserPromise.complete(fin)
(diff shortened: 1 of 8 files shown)


modules/core/src/weaver/MemoisedResource.scala
```diff
@@ -0,0 +1,65 @@
...
+        case Uninitialised =>
+          Uninitialised -> Concurrent[F].raiseError(
+            new IllegalStateException("Implementation error"))
+        case InUse(_, finaliser, n) if n <= 1 =>
+          Uninitialised -> finaliser
+        case InUse(value, finaliser, n) =>
+          InUse(value, finaliser, n - 1) -> Concurrent[F].unit
+      }.flatten
+
+      Resource.make(initialise)(_ => finalise)
```
RaasAhsan, line 62: Resource allocation and release are both uncancelable in CE2 so this is cancellation-safe

RaasAhsan: In CE3 the `make` constructor is also uncancelable, but the underlying `Resource` that is `allocated` may be cancelable so may be losing some cancellation timeliness by doing this. might be good to explore that further (if you even need it, that is :P )

RaasAhsan: Actually, the above is also true for CE2: you can cancel in between multiple `Resource` acquisitions, but not within a single acquisition. There's no point in trying to get around that in CE2 though :P

Baccata: That's great, thanks for having a look ! I'm gonna have a look at makeFull, it's probably worth specialising the implementations.

Because the construct is provided for the user to share resources between test suites (via indirection), I'd rather ensure that it's cancellation-safe, as I have no clue what people might want to be doing with this capability. 

 I might punt that to a later point though.

PR #215, opened 2021-03-01
Add scalatest interop module

dabd:
A scalatest interop module with scalatest's [must](https://www.scalatest.org/scaladoc/3.2.5/org/scalatest/matchers/must/index.html) and [should](https://www.scalatest.org/scaladoc/3.2.5/org/scalatest/matchers/should/index.html) matchers.

### DIFF
--- a/modules/scalatest/src/weaver/scalatestcompat/Matchers.scala
+++ b/modules/scalatest/src/weaver/scalatestcompat/Matchers.scala
@@ -1,0 +1,27 @@
+package weaver.scalatestcompat
+
+import cats.Monoid
+import weaver.{ EffectSuite, Expectations, SourceLocation }
+import weaver.Expectations.Helpers._
+import org.scalatest.Assertion
+import scala.util.{ Failure, Success, Try }
+
+trait Matchers[F[_]] {
+  self: EffectSuite[F] =>
+
+  implicit def toExpectations(
+      a: Assertion
+  )(
+      implicit pos: SourceLocation
+  ): Expectations =
+    Try(a) match {
+      case Success(_) => Monoid[Expectations].empty
+      case Failure(t) => failure(t.getMessage)
+    }
+
+  implicit def toExpectationsF(
+      a: Assertion
+  )(
+      implicit pos: SourceLocation
+  ): F[Expectations] = effectCompat.effect.pure(toExpectations(a))
+}
--- a/README.md
+++ b/README.md
@@ -41,6 +41,8 @@ libraryDependencies +=  "com.disneystreaming" %% "weaver-scalacheck" % "x.y.z" %
 // optionally (for specs2 interop)
 libraryDependencies +=  "com.disneystreaming" %% "weaver-specs2" % "x.y.z" % Test
 
+// optionally (for scalatest interop)
+libraryDependencies +=  "com.disneystreaming" %% "weaver-scalatest" % "x.y.z" % Test
 ```
 
 ## Motivation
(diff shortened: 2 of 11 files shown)


modules/scalatest/src/weaver/scalatestcompat/Matchers.scala
```diff
@@ -0,0 +1,27 @@
...
+  self: EffectSuite[F] =>
+
+  implicit def toExpectations(
+      a: Assertion
+  )(
+      implicit pos: SourceLocation
+  ): Expectations =
+    Try(a) match {
+      case Success(_) => Monoid[Expectations].empty
+      case Failure(t) => failure(t.getMessage)
```
dabd, line 19: scalatest throws a [TestFailedException](https://www.scalatest.org/scaladoc/3.2.5/org/scalatest/exceptions/TestFailedException.html) to indicate that a test failed.  Here we fail on any exception. Comments?

modules/scalatest/src/weaver/scalatestcompat/Matchers.scala
```diff
@@ -0,0 +1,27 @@
...
+import weaver.{ EffectSuite, Expectations, SourceLocation }
+import weaver.Expectations.Helpers._
+import org.scalatest.Assertion
+import scala.util.{ Failure, Success, Try }
+
+trait Matchers[F[_]] {
+  self: EffectSuite[F] =>
+
+  implicit def toExpectations(
+      a: Assertion
```
Baccata, line 13: The big problem of scalatest's assertions is that they are throwing, so this doesn't quite work because by the time the value should be passed to this function, it'd already have thrown on a failed case. Which means this parameter ought to be passed by-name. 

This also indicates a larger problem : throwing exceptions are bound to break the principle of least surprise, because you can't trust that the implicit conversion will kick-in in all cases. The reason I've accepted a specs2 integration module is because assertions in specs2 are pure values, which are inherently safer when it comes to refactoring. 

Here's a concrete example of what's gonna be problematic with a scalatest integration : 

```scala
val foo = (1 mustBe 2) || (2 mustBe 2) 
```

This will not throw, because the compiler will see `||` being used and will understand that it must implicitly convert to something that supports `||`. It'll find a by-name conversion to `Expectations` and use it. However, let's assume the following refactoring, which the user might rightfully expect to mean the exact same thing : 

```scala 
val foo1 = (1 mustBe 2)
val foo2 = (2 mustBe 2) 
val foo = foo1 || foo2 
```

foo1 will not be converted to Expectations, because the compiler has no indication that it should be the case. Therefore, that line will throw and the test will fail, although the user could expect it to succeed.  

Because of this, which is surprising behaviour, I'd rather not accept a scalatest integration in this form.

dabd: I think we would have an additional problem with by-name implicit arguments in an implicit conversion: https://github.com/scala/scala/pull/8590.  If I understood correcly, if given a block (multiline expression), the compiler will use only the result (last value) to kick-in the implicit def and not the whole block.

Baccata: That is also true. Which is less problematic when you work with pure values (which scalatests assertions are not)

dabd: @Baccata Do you propose any workaround to add a scalatest integration?

Baccata:
I cannot propose a simple workaround at this time, I'm afraid. Despite the fundamental problem of scalatest's assertions that don't play well with the paradigm weaver is abiding by, the PR did look good.

PR #228, opened 2021-03-23
Discipline integration

keynmol:
@Baccata this turned out a bit hairier than I expected, because `EffectSuiteAux` is barren and has no way of registering tests.

TODO:

- [x] Figure out whatever happened to scalafix on CI
- [x] Add TestParameters handling to `checkAll`


modules/discipline/src/weaver/discipline/Discipline.scala
```diff
@@ -0,0 +1,49 @@
...
+        }
+    }
+}
+
+object Discipline {
+
+  private[discipline] case class DisciplineException(
+      input: List[Arg[Any]],
+      cause: Throwable)
+      extends Exception(cause)
```
Baccata, line 39: I think we have an exception for that

modules/discipline/src/weaver/discipline/Discipline.scala
```diff
@@ -0,0 +1,49 @@
...
+import scala.util.control.NoStackTrace
+
+import org.scalacheck.Test.{ Passed, Proved, Exhausted, Failed, PropException }
+
+trait Discipline { self: EffectSuiteAux =>
+
+  import Expectations.Helpers._
+  import Discipline._
+
+  def checkAll(name: String, ruleSet: Laws#RuleSet): Unit =
```
Baccata, line 18: what if we returned `Expectations` instead ? That'd solve the problem of not having "pureTest" accessible, and would align with what we do in other integrations

docs/discipline.md
```diff
@@ -0,0 +1,46 @@
...
+import weaver._
+import weaver.discipline._
+import cats.kernel.laws.discipline.EqTests
+
+object DisciplineTests extends SimpleIOSuite with Discipline {
+  
+  override def maxParallelism = 1
+
+  checkAll("Int", EqTests[Int].eqv)
+  checkAll("Boolean", EqTests[Int].eqv)
```
Baccata, line 40: EqTests[Boolean] ?

modules/discipline/src/weaver/discipline/Discipline.scala
keynmol: I think this would offer a worse experience than similar integrations - right now we get a test per rule, and if we collapse everything into expectations, it'd be hard to pick out the particular law violated

image

We could alternatively just make it depend on weaver-cats :shrug: 

I'll see what it looks like if we collapse all rules into 1 value

Baccata: > right now we get a test per rule, and if we collapse everything into expectations, it'd be hard to pick out the particular law violated 

It wouldn't, since it'd essentially equate to a foldMap (failed expectations are reported individually). You just wouldn't see one line per passed law 

The problem with the current state of the PR is is that the change in the API forces a mutation-like way of adding tests in an abstract construct where that decision has not yet been made, and that is not something I'm willing to accept easily, even if this suite style is the one we advertise most prominently across the codebase. 

So I see two ways of solving this : 

* make "checkAll" return Expectations, as pointed out previously 
* having a `DisciplineSuite` instead, that'd take the law set as a parameter, so 1 law set == 1 suite. The idea of having "pure suites" that wouldn't require effect types was evoked previously, and could be leveraged so that users wouldn't have to ascribe a particular effect when creating a `DisciplineSuite`, and it'd be generally more suited for unit-tests. 

NB : the two solutions could actually be provided

keynmol: I'm not a fan of extending the core trait either, especially for such a simple integration.

How about we put a `PureSuite` in core with a single test builder (equivalent to `pureTest`), with no mention of effects and just funnel it through a pure fs2 runner?

I think this can also help people that now need to "know" that you need to set `maxParallelism ` to 1 to avoid the overhead of parallel runner in pure tests?

Baccata: > How about we put a PureSuite in core with a single test builder (equivalent to pureTest), with no mention of effects and just funnel it through a pure fs2 runner?

Yeah I think that's what I'd like for us to be aiming for. I'd probably call it "FunSuite" rather than "Pure", since I find it somewhat ironic that "PureSuite" would construct its tests using a mutative function. 

It'd have to be a preliminary PR as it's probably gonna be harder than it sounds (due to lack of covariance on the "Suite" trait). Might be mistaken though

docs/funsuite.md
```diff
@@ -0,0 +1,41 @@
+---
+id: funsuite
+title: Pure tests
+---
+
+If your tests don't perform and effects, you can use a simplified interface `FunSuite`,
```
Baccata, line 6:
```suggestion
If your tests do not require any capability provided by effect-types, you can use a simplified interface `FunSuite`,
```

PR #243, opened 2021-04-02
Pure suite

keynmol:
- Adds `FunSuite` with a single `test` method for pure tests.
- Intended to avoid the overhead of running pure tests through IO and make integrations with things like discipline easier (https://github.com/disneystreaming/weaver-test/pull/228/files#r599878255)

TODO:

- [ ] make sure this is the fastest way of running pure tests
- [ ] add documentation
- [x] ~May be add `FunSuite` as top level in each of the frameworks? `object Tests extends weaver.FunSuite` can work everywhere the same way~ 
    - Not such a good idea, probably, in case the user is mixing several weaver frameworks in one project (is that even going to work?)


Baccata:
I'm fine with making these FunSuites effect specific, even if no effect is actually ran. It means that in the eventually that people would have several of our frameworks in the same classpath, the funSuites would still be ran only once, and I don't think we gain much from trying to make them 100% agnostic wrt effect types.

Baccata (approved):

Baccata:
Merged to facilitate the Junit work , docs can come in a later PR

PR #250, opened 2021-04-06
Scala 3.0.0-RC2

keynmol:


Baccata (approved):

keynmol:
1 more change - I'll change the job name to `3.0` given that we only publish against single scala 3 version

Issue #280, opened 2021-04-24
specs2 integration - using `and` on `beSome`'s `ValueCheck` leads to failure


LLCampos:
WeaverVersion = 0.7.0
Specs2Version = 4.10.6

The following passes:

```scala
import org.specs2.mutable._

class MySuite extends Specification {
  "my test" >> {
    Some(1) must beSome((i: Int) => {
      (true must beTrue) and (true must beTrue)
    })
  }
}
```

But the equivalent in weaver doesn't.

```scala
import weaver._
import weaver.specs2compat.IOMatchers

object MySuite2 extends SimpleIOSuite with IOMatchers {
  test("my test") {
    Some(1) must beSome((i: Int) => {
      (true must beTrue) and (true must beTrue)
    })
  }
}
```

```
MySuite2
- my test 19ms
*************FAILURES**************
MySuite2
- my test 19ms
  Some(1) is Some but 1: java.lang.Integer != MySuite2$$$Lambda$175/352076319@76ab2cc5: MySuite2$$$Lambda$175/352076319 (../experiment-with-stuff-scala/src/test/scala/weaver.scala:6)

Execution took 19ms
1 tests, 1 failed
```

LLCampos:
I'm looking into this ^^

LLCampos:
`beSome` expects a function that returns a `MatchResult`, but when we use `and` we force specs2's `MatchResult` into weaver's` Expectations`. specs2 has its own `and` but it's overloaded by weaver's.

This means that if, inside `beSome` we use any function that returns `Expectations` (e.g., weaver's `and`, `or`, `xor`) the test will fail.

I added some failing tests [here](https://github.com/disneystreaming/weaver-test/pull/281).

Baccata:
Ergh that's a pickle. To solve this in userland, rather than using `IOMatchers`, you could override the `test` method (which provides a few alternative shapes by means of a PartiallyApplied construct) and add some additional shapes that'd take the same thing but `MatchResult` instead, and would delegate to the original methods after having transformed the MatchResult to Expectations. 

Since IOMatchers is meant to be mixed in with other traits, I don't think we can use that technique upstream ... I guess we could have a `spec(...)` method that'd do the same thing as `test` though, with matchers instead of expectations. 

What do you think ?

LLCampos:
@Baccata, I would say that having a different a different test method (`spec(...)`) is undesirable from a user experience point of view. :confused: Although it might be the best option if no alternative. 

[...]

PR #281, opened 2021-04-24
specs2 integration - Fix "using `and` on `beSome`'s `ValueCheck` leads to failure"

LLCampos:
# Summary of changes
## Major
* See https://github.com/disneystreaming/weaver-test/issues/280. Some `specs2` matchers expect a `MatchResult`/`Result` so we might have to convert them back from weaver's `Expectations`. This PR adds implicits that deal with those conversions.

## Minor
* Remove duplicated tests

### DIFF
--- a/modules/specs2/src/weaver/specs2compat/Matchers.scala
+++ b/modules/specs2/src/weaver/specs2compat/Matchers.scala
@@ -7,8 +7,8 @@ import cats.effect.IO
 
 import weaver.{ AssertionException, EffectSuite, Expectations, SourceLocation }
 
-import org.specs2.execute.{Failure, Result, Success}
-import org.specs2.matcher.{MatchResult, MustMatchers, StandardMatchResults}
+import org.specs2.execute.{ Failure, Result, Success }
+import org.specs2.matcher.{ MatchResult, MustMatchers, StandardMatchResults }
 
 trait Matchers[F[_]] extends MustMatchers {
   self: EffectSuite[F] =>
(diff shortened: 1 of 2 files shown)


Baccata:
I have a rather busy week and can't review straight away, so thank you in advance for your patience.

Baccata:
Alright so I've given it some thought, I'm not too keen on merging this at this time: implicit conversions have proven brittle  and have bitten our users in unfortunate ways (including yourself 😉 ) 

I think moving forward they should be avoided (we've already taken steps to remove a number of them in this codebase), and there are ways to do so using polymorphism and typeclasses ... however the main annoyance is the fact that the source location which (Expectations require) is provided via macro, and that in the current state of the language it's next to impossible to capture them against a `MatchResult` without relying on implicit conversions. 

@LLCampos, what do you value most between the ergonomics and source-location reporting ? The way I see it right now, it could go two ways : 
1. remove implicit conversions altogether and rely and explicit `.toExpectation` extension method (provided via implicit class in scala2) 
2. ensure that the `test` could take specs2's MatchResults, but the reported location would be the one of the `test` call (as opposed to the matcher call). 

I'm trying to think of a more general solution to this problem. There might be one, but it's the kind that would likely justify a minor-version bump, and we're hoping to avoid it at least until the Scala 3.0 release...

LLCampos:
Right. :)

Erm, personally, if in one of my projects I had to add `.toExpectation` to everything, I think I would rather just throw away the `specs2` matchers and use the default `weaver`'s tooling.

> 2. ensure that the test could take specs2's MatchResults, but the reported location would be the one of the test call (as opposed to the matcher call).

What would this mean in practice? If the test fails, what does the user see?

Baccata:
> What would this mean in practice? If the test fails, what does the user see?

Exactly what they see now I think, except that the reported line number would indicate where `test("foo")` is as opposed to `x mustEqual y` is.

LLCampos:
Right :) That's what I had in mind but wanted to clarify.

Well, people seem to be happy be using Doobie/fs2/cats-effect with the less than perfect error messages source-location-wise.

I think if this trade-off is documented in the same place where it is explained how to integrate with the specs2 matchers, so that people can do an informed decision, it should be ok?

Baccata:
We're plaining and simply removing support for the specs2 matchers, to ease maintenance burden. Apologies for the inconvenience.

PR #283, opened 2021-04-25
Keep result of any task in `UnsafeRun.sync`

kubukoz:
This should allow users to `unsafeRun` a task without providing an `IORuntime` themselves, e.g. if they want to add some customization layer between weaver and their tests.

I was hoping for binary compatibility, but `void` and `Object` probably don't play well...

Update: now I noticed weaver is using `global` 😅 what could go wrong...

### DIFF
--- a/modules/core/cats/src-ce2/weaver/CatsUnsafeRun.scala
+++ b/modules/core/cats/src-ce2/weaver/CatsUnsafeRun.scala
@@ -23,7 +23,7 @@ trait CatsUnsafeRun extends UnsafeRun[IO] {
 
   def cancel(token: CancelToken): Unit = sync(token)
 
-  def sync(task: IO[Unit]): Unit = task.unsafeRunSync()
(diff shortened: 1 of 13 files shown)


Baccata:
I'm think we didn't intend for this method to be public 😅 we haven't cleaned house yet on the account of being mostly two nerds who've dealt with an enormous amount of churn in the upstream libraries :trollface: and haven't had much time to properly scrub the kitchen. 

Anyhow, can you elaborate on the usecase you're thinking of using it for ? The framework uses it for running individual suites from IDEs and that's pretty much it, and I'm wary of promising to keep a method that'd let users happily call unsafeRunSync outside of a Dispatcher (I trust you with how you'd use this, I wouldn't trust 90% of people). 

> Update: now I noticed weaver is using global 

Talking about CE2 or CE3 there ? Also, overriding is possible (not trivial but definitely not hard either).

kubukoz:
Closing for age, I don't remember what the actual usecase even was.

PR #302, opened 2021-05-16
Scala 3 final and group dependency versions

keynmol:
I've just published Expecty 0.15.3, but it will take time to propagate.

Also grouped the dependencies together as they're all over the place.

Plan is to publish this as 0.6.3 before we do a 0.7.0/0.8.0 final.


keynmol:
Right. My half-arsed logic for parsing project names breaks because `CE3` and `3` (Scala3 binary suffix) both end in `3` :D 

I'm going to re-implement it with https://github.com/indoorvivants/sbt-commandmatrix and see if that improves things.

project/WeaverPlugin.scala
```diff
@@ -429,14 +429,15 @@ object WeaverPlugin extends AutoPlugin {
           var projectId = lp.project
 
           val scalaAxis =
-            if (projectId.endsWith(scala3Suffix)) {
+            if (projectId.endsWith(scala3Suffix) && !projectId.endsWith(ce3Suffix)) {
```
keynmol, line 432: This is ugly and horrible, and commandmatrix was going to make this better, until I realise there's some weird non-lazy initialisation going on somewhere :D 

So it will have to wait.

PR #306, opened 2021-05-19
broken: Attempt to publish for ZIO on Scala3

keynmol:
I thought it was going to be easy-peasy-lemon-squeezy, but it turned out difficult-difficult-zio-difficult


Baccata (approved):

keynmol:
Turns out this is due to the same bug we have in Expecty: https://github.com/eed3si9n/expecty/issues/54

Running with a locally published version fixes it.

PR #318, opened 2021-06-10
Improve assertion docs

keynmol:
1) TL;DR of expectations on top
2) Detailed examples showing the output/purpose/best practices of various expectations
3) Adding suite outputs to some other pages

See the full result here: http://weaver-site.surge.sh/weaver-test/docs/expectations


Baccata:
That looks really neat !

Baccata (approved):

docs/resources.md
```diff
@@ -45,3 +45,6 @@ object HttpSuite extends IOSuite {
 }
 ```
 
+```scala mdoc:passthrough
+println(weaver.docs.Output.runSuites(HttpSuite).unsafeRunSync())
```
keynmol, line 49: One worry is that this test suite actually runs an HTTP request to httpbin...

I'll probably revert this, don't want it to break during CI.

modules/docs/src/main/scala/weaver/FansiModifier.scala
```diff
@@ -0,0 +1,37 @@
+package weaver
+package docs
+
+import mdoc._
+
+class FansiModifier extends mdoc.PostModifier {
```
keynmol, line 6: This is not actually used, but I struggle to part ways with it, given how ugly of a hack it is.

website/static/css/custom.css
```diff
@@ -49,4 +49,5 @@ div.terminal pre {
   background-color: black;
   color: white;
   padding: 20px;
+  overflow: auto
```
keynmol, line 52: That's for the ScalaCheck integration page: http://weaver-site.surge.sh/weaver-test/docs/scalacheck 

The error message is hella long and breaks rendering.

docs/resources.md
keynmol: Changed in 3fd0aee, arguably better example as a whole: http://weaver-site.surge.sh/weaver-test/docs/resources

PR #331, opened 2021-07-07
ForEachTest resource

amumurst:
Add IOForEachSuite as a to acquire a new instance of a resource for each test in a suite.

Combining this with global makes it possible to do things that I have not seen in other test frameworks, for instance having the global scope setting up a Database provider and shutting it safely down after all tests are done, but still allowing tests to be written across multiple files, without dependencies on each other and with every test getting a fresh database. 

I have tried to implement and test this pattern in `UniqueResourceSuite` and `SharedResourceSuite`.

If this is something you want to move forward with I would be happy to look at adding it to the docs.

### DIFF
--- a/modules/core/src/weaver/suites.scala
+++ b/modules/core/src/weaver/suites.scala
@@ -58,10 +58,9 @@ abstract class RunnableSuite[F[_]] extends EffectSuite[F] {
     effectCompat.sync(run(args)(outcome => effectCompat.effect.delay(report(outcome))))
 }
 
-abstract class MutableFSuite[F[_]] extends RunnableSuite[F]  {
+abstract class MutableBaseFSuite[F[_]] extends RunnableSuite[F]  {
 
   type Res
-  def sharedResource : Resource[F, Res]
 
   def maxParallelism : Int = 10000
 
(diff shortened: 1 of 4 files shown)


modules/core/src/weaver/suites.scala
```diff
@@ -81,37 +80,63 @@ abstract class MutableFSuite[F[_]] extends RunnableSuite[F]  {
...
+
+  private[this] var isInitialized = false
+
+  private[this] def initError() =
+    new AssertionError(
+      "Cannot define new tests after TestSuite was initialized"
+    )
+
+
+  private[weaver] def getTests(args: List[String]) = {
```
amumurst, line 95: This stuff is just abstracted back up to the common base trait between forEach and standard suite, no changes made.

Baccata:
@amumurst hi, and thanks for the contribution ! This feature had been requested by several people and was on my backlog, so the PR makes me happy. 

I'm gonna take some time (probably tomorrow) to give it a proper read and reflect about the implications of your implementation, but at the very least I can tell you that there is interest 😄

modules/core/src/weaver/suites.scala
```diff
@@ -58,10 +58,9 @@ abstract class RunnableSuite[F[_]] extends EffectSuite[F] {
     effectCompat.sync(run(args)(outcome => effectCompat.effect.delay(report(outcome))))
 }
 
-abstract class MutableFSuite[F[_]] extends RunnableSuite[F]  {
+abstract class MutableBaseFSuite[F[_]] extends RunnableSuite[F]  {
```
Baccata, line 61: Okay, so I think you're on the right track, however what I'd like to see is this : 

```scala
trait MutableBaseFSuite[F[_]] {
    type Res 
    // PerTestRes would be replace Res in the "test" methods 
    type PerTestRes 

    def sharedResource: Resource[F, Res] 
    def perTestResource(res: Res) : Resource[F, PerTestRes] 
}
```

and then MutableFSuite would become :  

```scala
trait MutableFSuite[F[_]] extends MutableBaseFSuite[F] {
   type PerTestRes = Res
   final def perTestResource(res: Res) : Resource[F, Res] = Resource.pure(res)  
}
```

The reason for doing it this way is that : 
1. it works for your usecase 
2. people can still share resources a little bit more granularly than your current version, which is "shared XOR per test", whereas this version would be "shared OR per test" 
3. it works on the javascript backend where cross-suite sharing is not possible, but per suite sharing is possible
4. You can still have your `MutableForEachSuite` in userland : 

```scala
trait MutableForEachSuite[F[_]] extends MutableBaseFSuite[F] {
   type Res = Unit 
   def uniqueResource : Resource[F, PerTestRes]
   final def perTestResource(res: Unit) : Resource[F, PerTestRes] = uniqueResource
}  
```

Baccata: @amumurst, just to touch base : what do you think about the suggestion above ?

vladimir-popov:
Any plans to bring this feature to the project?

Issue #338, opened 2021-07-16
Display initial seed in property-check failures ...


Baccata:
... (as opposed to "current" seed, which doesn't quite work when parallelism is involved due to non-determinism) 

Also change the config to take a "Seed" as opposed to a long.

PR #339, opened 2021-07-20
Fix #337 and #338 

### LINKED ISSUE #337: Paths are not correctly relativised when using bloop
Such things get rendered : ` assertion failed (../../../bar/src/test/scala/Foo.scala:13)`

Baccata:
This attempts to parameterise the displayed source path based on the compiler settings, as opposed to using "pwd", which doesn't quite work with bloop as the server will give whatever directory it was started in. 

Right now it is impossible to extract the workspace information from the macro context in Scala 3 (as it possible is in Scala 2) 

David Barri's got a PR in flight to add some API to retrieve macro settings (see https://github.com/lampepfl/dotty/pull/12039), but that's limited to a macro-specific compiler setting, whereas we'd need to access the full compiler-settings to query the workspace from the compiler. 

Let's see how that PR flies with the scala 3 maintainers, and whether we can contribute some api that'd let us query the whole set of compiler settings.

### DIFF
--- a/build.sbt
+++ b/build.sbt
@@ -42,11 +42,7 @@ val Version = new {
   }
 
   object CE2 {
-<<<<<<< HEAD
     val fs2        = "2.5.9"
-=======
-    val fs2        = "2.5.8"
->>>>>>> fe2b288... Remove scoverage, bump versions
     val cats       = "2.5.1"
     val zioInterop = "2.5.1.0"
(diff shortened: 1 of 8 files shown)


keynmol:
I don't think either of us remember what this is, so will close for inactivity - let's reopen at the strike of inspiration

Issue #387, opened 2021-09-16
Run tests with a compound effect


iRevive:
Scala version: `3.0.1`, `2.13.6`
Weaver version: `0.7.4`
Weaver framework (runner): `weaver.framework.CatsEffect`

Example repository: https://github.com/iRevive/weaver-effect-runner

## Synopsis 

Currently, there is no simple way to run tests with the compound effect. 
A test suite must extend `weaver.BaseCatsSuite`. Otherwise, the suite is not discovered by the test framework. 

## Example

Effect definition:
```scala
final case class RaisedError(error: Throwable, pos: Position, errorId: String)
final case class LogContext(traceId: String, extra: Map[String, String])

type WithLogContext[F[_], A] = Kleisli[F, LogContext, A]
type WithError[A]            = EitherT[IO, RaisedError, A]
type Eff[A]                  = WithLogContext[WithError, A]
```

Resource definition:
```scala
final case class App[F[_]](service: Service[F])

object App {
  // emulate allocation of the app
  def create[F[_]: Sync]: Resource[F, App[F]] =
    Resource.pure(new App[F](new Service[F]))
}

class Service[F[_]: Sync] {
  def multiply(a: Int, b: Int): F[Int] =
    Sync[F].delay(a * b)
}
```

Test case definition:
```scala
trait AppSuite extends MutableFSuite[Eff] with Expectations.Helpers with Checkers {
   type Res = App[Eff]
   def sharedResource: Resource[Eff, App[Eff]] = App.create[Eff] // resource allocation
   protected implicit def effectCompat: UnsafeRun = new EffUnsafeRun(name) // provide effect runner
}

object ServiceSuite extends WithAppSuite {
  test("multiply correctly") { app => ...  }
}
```

The suite is not discovered:
```
$ sbt show definedTests
[info] compiling 2 Scala sources to ${BASE}/weaver-effect-example/target/scala-3.0.1/classes ...
[info] *
```

## Workaround

Defining an extra suite extended from `CatsSuite` solves the problem.

```scala
trait WithAppSuite extends MutableFSuite[Eff] with Expectations.Helpers with Checkers { self =>
  type Res = App[Eff]
  def sharedResource: Resource[Eff, App[Eff]] = App.create[Eff]
  protected implicit def effectCompat: EffUnsafeRun = new EffUnsafeRun(name)

  abstract class IOSuite extends SimpleMutableIOSuite {
[...]

PR #388, opened 2021-09-16
Play with "provider"-fingerprinted construct

Baccata:
This should allow for some flexibility in the effects that suites can use whilst working against a given framework.

See https://github.com/disneystreaming/weaver-test/issues/387

### DIFF
--- a/modules/core/cats/src-ce2/weaver/BaseIOSuites.scala
+++ b/modules/core/cats/src-ce2/weaver/BaseIOSuites.scala
@@ -8,9 +8,11 @@ trait BaseIOSuite extends BaseCatsSuite { self: RunnableSuite[IO] =>
   final implicit protected def contextShift: ContextShift[IO] =
     effectCompat.contextShift
   final implicit protected def timer: Timer[IO] = effectCompat.timer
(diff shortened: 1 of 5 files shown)


Baccata:
@keynmol could you give this a review when you have time ? I'd like to know what you think

keynmol:
I * think* it's fine. 

- It doesn't seem to break the hierarchy in integrations?
- Does `EffectSuite` need to extend Provider itself, or do we expect any user of EffectSuite directly to provide `getSuite`? 
- I would like to run this branch on
  - our internal weaver extensions module
  - internal set of tests to make sure this doesn't affect discovery in some weird and interesting way

Baccata:
> Does EffectSuite need to extend Provider itself, or do we expect any user of EffectSuite directly to provide getSuite

the latter. If EffectSuite was to extend Provider directly, it wouldn't solve the problem, as an `EffectSuite[OptionT[IO, *]]` could not also inherit from `EffectSuite.Provider[IO]` because of shared erasure. 

> I would like to run this branch on [...] 

That's fine by me 👍

Baccata:
@iRevive apologies for the time it took for us to merge this, work's been quite busy

PR #393, opened 2021-09-20
EffectProvider. Fix compilation issues

iRevive:
I fixed compilation issues in the following modules: 
- monix
- monix-bio
- zio

### DIFF
--- a/modules/core/monix/src/weaver/monixcompat/suites.scala
+++ b/modules/core/monix/src/weaver/monixcompat/suites.scala
@@ -6,7 +6,7 @@ import cats.effect.Resource
 import monix.eval.Task
 import monix.execution.Scheduler
 
-trait BaseTaskSuite extends EffectSuite[Task]
+trait BaseTaskSuite extends EffectSuite.Provider[Task]
 
 abstract class PureTaskSuite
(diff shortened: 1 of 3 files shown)


Baccata:
Outstanding, thanks ! 

@iRevive Since you're the first person asking for this feature, what would be your expectations when writing a suite that'd use [global resources](https://disneystreaming.github.io/weaver-test/docs/global_resources). The tricky part is that because of how global resources are injected in the constructor of the suites, we cannot retrieve the effect information from the suite directly (since it's not instantiated when we need the information). 

So as it stands on this branch, if you need global resources, you'll have to manually lift the IO resources into your effect type. Do you deem it acceptable ?

iRevive:
@Baccata I do not use GlobalResouce in my projects, but from my point of view, it's more than fine to lift IO resources into a custom effect type.

I need to run a test suite with contextual effect (Kleisli + IO) to achieve several goals:
- ad-hoc configuration of Jaeger spans
- ad-hoc replacement of `io.odin.Logger` with `weaver.Log` to print non-error log statements (debug, info) of failed tests

If there is a shared resource, which is used by multiple test suites, the allocation with contextual effect does not make a lot of sense to me.

PR #404, opened 2021-10-13
#400 add alias for `apply` in `PartiallyAppliedTest`

bwiercinski:
resolves https://github.com/disneystreaming/weaver-test/issues/400

I'm more than welcome to hear other candidates for the name. I was also considering names: `match_`, `matching`, `usingResource`, `usingSharedResource`, but I like short names.


modules/core/src/weaver/suites.scala
```diff
@@ -79,6 +79,9 @@ abstract class MutableFSuite[F[_]] extends RunnableSuite[F]  {
     def apply(run: => F[Expectations]) : Unit = registerTest(name)(_ => Test(name.name, run))
     def apply(run : Res => F[Expectations]) : Unit = registerTest(name)(res => Test(name.name, run(res)))
     def apply(run : (Res, Log[F]) => F[Expectations]) : Unit = registerTest(name)(res => Test(name.name, log => run(res, log)))
+
+    // this alias helps using pattern matching on `Res`
+    def useRes(run : Res => F[Expectations]) : Unit = apply(run)
```
Baccata, line 84:
```suggestion
    def usingRes(run : Res => F[Expectations]) : Unit = apply(run)
```

Baccata: pretty much yeah

PR #419, opened 2021-11-04
Clarify plugins required for intellij support

lukestephenson:
I hadn't needed this plugin before doing Scala development with ScalaTest.  So I needed to track down this needed adding.


Baccata:
Thanks for this !

PR #432, opened 2021-11-29
[Tracking] Scala 3.1.0

keynmol:
FS2 3.2.2 and discipline 1.3.0 bring in Scala 3.1.0, which would force us to upgrade.

Opening this PR to track and redirect all the Steward PRs that break because of 3.1.0.

Given CE didn't pull the trigger on Scala 3.1.0 (at least CE 3.3.0 didn't), I think we should do it at some point (as we're not at the root of people's dependency tree, but near the leaves)

### DIFF
--- a/modules/core/src-scala-3/SourceLocationMacro.scala
+++ b/modules/core/src-scala-3/SourceLocationMacro.scala
@@ -25,7 +25,7 @@ object macros {
 
     val position = Position.ofMacroExpansion
 
-    val psj = position.sourceFile.jpath
+    val psj = position.sourceFile.getJPath.get
     // Comparing roots to workaround a Windows-specific behaviour
     // https://github.com/disneystreaming/weaver-test/issues/364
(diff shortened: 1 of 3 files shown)


modules/core/src-scala-3/SourceLocationMacro.scala
```diff
@@ -25,7 +25,7 @@ object macros {
 
     val position = Position.ofMacroExpansion
 
-    val psj = position.sourceFile.jpath
+    val psj = position.sourceFile.getJPath.get
```
keynmol, line 28: Source: https://github.com/lampepfl/dotty/blob/7e0689996e0fd62012ba4982a4a6aefa23518c66/library/src/scala/quoted/Quotes.scala#L4240-L4247

I don't think `.get` makes this less safe, as previously it could return null (for virtual files) and would blow up anyways.

May be we need to turn SourceLocation into ADT, but I worry it's a breaking change for very little benefit.

PR #477, opened 2022-02-16
Wait for resources release before ending tests

### LINKED ISSUE #473: GlobalResource's do not get released while using forkedJvm in SBT
When using multiple `GlobalResource's` in a `forked` JVM in SBT, not all the resources are released, i.e. release in `Resource.make(acquire)(release)` never gets invoked.

Frequently only the first GlobalResource is released gracefully.

I suspect this is because Weaver attempts to close the resources, and in parallel, sends the test outcome to SBT. Once SBT gets the test outcome, it terminates the forked JVM before all resources have the release task executed on them.

If not using a forked jvm you can see the `Resource's` are released after the tests have completed in SBT.

I would expect the behavior to be all Resource's are released before signaling to SBT the tests/task is complete.

If we use a non global shared resource, resource's are released as expected.

I have a sample project at https://github.com/ssahayam-zd/weaver-resource-cleanup which demonstrates this.

lukestephenson:
Fixes #473

A java Semaphore is created initially with 0 permits.  An additional cleanup Resource
is composed around the resources required by the tests.  When the test Resources
complete, the clean Resource releases a permit on the Semaphore.

SBTTask waits for this Semaphore to aquire a permit before wrapping things up.

Kudos to @Baccata for help with the fix.

### DIFF
--- a/modules/framework/src-jvm/RunnerCompat.scala
+++ b/modules/framework/src-jvm/RunnerCompat.scala
@@ -45,13 +45,18 @@ trait RunnerCompat[F[_]] { self: sbt.testing.Runner =>
 
   private def runBackground(
       globalResources: List[GlobalResourceF[F]],
+      waitForResourcesShutdown: java.util.concurrent.Semaphore,
       tasks: List[IOTask],
       gate: Promise[Unit]): Unit = {
-    cancelToken = Some(unsafeRun.background(run(globalResources, tasks, gate)))
+    cancelToken = Some(unsafeRun.background(run(globalResources,
+                                                waitForResourcesShutdown,
+                                                tasks,
+                                                gate)))
   }
 
   def tasks(taskDefs: Array[TaskDef]): Array[Task] = {
-    val stillRunning = new AtomicInteger(0)
+    val stillRunning             = new AtomicInteger(0)
+    val waitForResourcesShutdown = new java.util.concurrent.Semaphore(0)
 
     val tasksAndSuites = taskDefs.toList.map { taskDef =>
       taskDef -> suiteLoader(taskDef)
@@ -86,6 +91,7 @@ trait RunnerCompat[F[_]] { self: sbt.testing.Runner =>
         new SbtTask(taskDef,
                     isDone,
                     stillRunning,
+                    waitForResourcesShutdown,
                     promise,
                     queue,
                     loggerPermit,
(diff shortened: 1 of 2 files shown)


Baccata:
@lukestephenson, considering @aevanszen's handle, I presume you guys work together at Zendesk. 

Can you confirm that this patch fixes the issue ? Ideally we'd add some tests, but testing this stuff is somewhat tricky, and if the patch seems to solve the issue, I'm inclined to merge/release asap.

modules/framework/src-jvm/RunnerCompat.scala
```diff
@@ -140,7 +150,13 @@ trait RunnerCompat[F[_]] { self: sbt.testing.Runner =>
       }
     }
 
-    preventDeadlock(resourceMap(globalResources)).use { read =>
+    val cleanupResource = Resource.make(effect.unit)(_ =>
```
keynmol, line 153: I want to bikeshed a little here and use `Resource.onFinalize`, but we can do it later, as long as this PR confirms the fix

lukestephenson:
@Baccata I can confirm the fix works with the test project https://github.com/ssahayam-zd/weaver-resource-cleanup from the original issue is bumped to a local version of weaver published  (ie `sbt publishLocal`)  with these changes.

lukestephenson:
And you are correct.  I work with @aevanszen at Zendesk.  We have started to adopt weaver.

modules/framework/src-jvm/RunnerCompat.scala
lukestephenson: I've struggled to get the weaver codebase working nicely with intellij.  So I'm finding coding / refactoring a little more difficult that normal.  What IDE are you all using for weaver dev?

keynmol: I exclusively use Metals (with NeoVim, but most of my work on Weaver was done in VS Code with Metals).

Perhaps the issue in IJ is the fact that Weaver has an extremely large build matrix - we have lots of modules that overlap with each other in almost all sources. I remember IJ struggling a lot with that, as it doesn't fit in its model.

The experience in Metals should be nicer, but on the other hand it doesn't have as many refactorings as IJ :) Swings and roundabouts

keynmol: Not sure which exact issue I was tracking, but it's a known problem: https://youtrack.jetbrains.com/issue/SCL-19569

Baccata (approved):

Baccata:
Thank you @lukestephenson! I've triggered the release of 0.6.10/0.7.10.

modules/framework/src-jvm/RunnerCompat.scala
Baccata: In theory, Intellij in BSP should work okay. I think the standard mode makes a lot of assumption that we happen to break in this codebase. Not 100% sure

Issue #480, opened 2022-02-17
Using testOnly with multiple matches results in multiple shared resource instantiations


ssahayam-zd:
Hi,

I followed the recommendations for using `testOnly` as per [Regarding "testOnly" build tool commands](https://disneystreaming.github.io/weaver-test/docs/global_resources#regarding-testonly-build-tool-commands) which works as expected when `testOnly` matches a single test.

For example:

```
sbt testOnly *VerySpecificSuite
```

if however, I run `testOnly` with a slightly broader search as as:

```
sbt testOnly *SomeCommonPrefixSuite
```

then I can see that the shared resource used across these Suites is instantiated multiple times.

I have a small sample [repo](https://github.com/ssahayam-zd/weaver-testonly) that highlights the problem.

I am not sure if I am using this feature incorrectly or whether this is a bug/feature.

Thanks!

Baccata:
First of all, thank you for writing such a great issue, with a repro, this is great 😄 

I'm sorry to say it's a user error. We have a note in the docs you linked : 

``` 
pass your GlobalResource implementations to the testOnly command, alongside the suite you really want to run
```

So you need 

```
sbt testOnly *SharingSuite SharedResource
```

When you do 

``` 
sbt testOnly *SharingSuite
```

the `SharedResource` class is not matched by the glob pattern, so it doesn't passed to weaver by SBT, which means that you always hit the `case None => baseResources` [here(https://github.com/ssahayam-zd/weaver-testonly/blob/main/src/test/scala/acme/weaverissue/SharedResource.scala#L21), which means that the resource gets recreated by the suite. 

The thing with shared resources objects in weaver is that they are essentially considered as a test suite by the build tool, so when you call `testOnly`, you need to ensure they are passed, otherwise weaver has no way of knowing that they exist.

ssahayam-zd:
> First of all, thank you for writing such a great issue, with a repro, this is great 😄

Thanks! 😄 


Also glad this is not a bug! 😅 

It looks like you need the full package path to the shared resource when running it through `testOnly`:

```
testOnly *SharingSuite acme.weaverissue.SharedResources
```

I'm happy to submit a PR updating the docs with a simple example as above.

keynmol:
This seems like it can be closed, reopen if I'm misreading it :)

PR #483, opened 2022-02-18
Add example on how to run shared resources with `sbt testOnly`

ssahayam-zd:
Hi all,

I'm updated the docs with instructions on how to specify a shared resource to `sbt testOnly`. This was part of the reason for my raising [Using testOnly with multiple matches results in multiple shared resource instantiations](https://github.com/disneystreaming/weaver-test/issues/480) and I think it could help others in the future.

### DIFF
--- a/docs/global_resources.md
+++ b/docs/global_resources.md
@@ -87,6 +87,8 @@ Some build tools provide a "testOnly" (or equivalent) command that lets you test
 An example of how to do this:
 
 ```scala mdoc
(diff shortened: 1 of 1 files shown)


Baccata:
Oops, I approved a bit quickly.

Mdoc does not accept package declarations. We could put a // before it or something to make mdoc happy

Baccata (approved):

Baccata:
Thanks !

PR #491, opened 2022-02-25
Allows resource startup timeout to be configured

mcgizzle:
Configurable by a new system property

For test setups that do a lot of work, 120 seconds can be too short.

Open to suggestions for the name and how to handle the `IllegalArgumentException` if the key is empty.


modules/framework/src-jvm/RunnerCompat.scala
```diff
@@ -119,9 +120,15 @@ trait RunnerCompat[F[_]] { self: sbt.testing.Runner =>
                   gate)
     stillRunning.set(sbtTasks.size)
 
+    val maybeTimeout: Option[FiniteDuration] = Try(System.getProperty("weaver.test.startupTimeout"))
```
Baccata, line 123: pro-tip : you can use `sys.props` (which gives you a scala map of the available system properties)

Baccata:
@mcgizzle can you run `scalafmtAll` in sbt and push ?

Baccata:
Merging, will fmt myself :)

mcgizzle:
Thanks :) I'm on leave this week and don't have my laptop with me, sorry!

Issue #479, opened 2022-02-17
Unable to resolve GlobalResource defined in a submodule/project


aevanszen:
If you define a GlobalResource in a subproject/module for re-use, i.e. a `test-support-module`, the GlobalResource cannot be resolved/used in a test in another module. You will get the below sample output:

```
[info] MyTestSuite
[info] - Unexpected failure 0ms
[info] *************FAILURES**************
[info] MyTestSuite
[error] - Unexpected failure 0ms
[error]   GlobalResourceF$ResourceNotFound: Could not find a resource of type java.lang.String with label null
[error]
[error]   GlobalResourceF.scala:110    weaver.GlobalResourceF$Read#$anonfun$getOrFailR$1
[error]   Stream.scala:3475            fs2.Stream$#$anonfun$resourceWeak$3
[error]   Stream.scala:3475            fs2.Stream$#$anonfun$resourceWeak$3$adapted
[error]   Stream.scala:1191            fs2.Stream$#$anonfun$flatMap$1
[error]   Algebra.scala:620            fs2.internal.FreeC$#$anonfun$flatMapOutput$1
[error]   Algebra.scala:53             fs2.internal.FreeC$$anon$1#cont
[error]   Algebra.scala:240            fs2.internal.FreeC$ViewL$$anon$9$$anon$10#<init>
[error]   Algebra.scala:240            fs2.internal.FreeC$ViewL$$anon$9#cont
[error]   Algebra.scala:234            fs2.internal.FreeC$ViewL$$anon$8#next
[error]   Algebra.scala:475            fs2.internal.FreeC$#$anonfun$compile$8
[error]   CECompat.scala:51            eval @ weaver.CECompat#resourceLift
[error]   CompileScope.scala:413       map @ fs2.internal.CompileScope#interruptibleEval
[error]   Algebra.scala:503            flatMap @ fs2.internal.FreeC$#go$1
[error]   Algebra.scala:463            flatMap @ fs2.internal.FreeC$#$anonfun$compile$7
[error]   Algebra.scala:460            flatMap @ fs2.internal.FreeC$#go$1
[error]   Algebra.scala:436            flatMap @ fs2.internal.FreeC$#interruptGuard$1
[error]   Algebra.scala:436            flatMap @ fs2.internal.FreeC$#interruptGuard$1
[error]   CompileScope.scala:185       flatMap @ fs2.internal.CompileScope#$anonfun$acquireResource$4
[...]

PR #494, opened 2022-03-04
Allow loading GlobalResources from libraries on the classpath.

aevanszen:
As discussed in https://github.com/disneystreaming/weaver-test/issues/479.

This is a minimal implmentation to allow GlobalResources to be reused in a submodule / external jar across projects.

I've tested on a local project and it looks to work as expected.

### DIFF
--- a/modules/framework/src-jvm/RunnerCompat.scala
+++ b/modules/framework/src-jvm/RunnerCompat.scala
@@ -17,6 +17,9 @@ import sbt.testing.{ Task, TaskDef }
 import CECompat.Ref
 import CECompat.Semaphore
 
+import java.util.ServiceLoader
+import scala.collection.mutable.ListBuffer
+
 trait RunnerCompat[F[_]] { self: sbt.testing.Runner =>
 
   protected val suiteLoader: SuiteLoader[F]
@@ -106,9 +109,9 @@ trait RunnerCompat[F[_]] { self: sbt.testing.Runner =>
         makeTasks(taskDef, mkSuite)
     }.unzip
 
-    val globalResources = tasksAndSuites.collect {
+    val globalResources = (tasksAndSuites.collect {
       case (_, suiteLoader.GlobalResourcesRef(init)) => init
-    }.toList
+    }.toSet ++ spiGlobalResources).toList
 
     // Passing a promise to the FP side that needs to be fulfilled
     // when the global resources have been allocated.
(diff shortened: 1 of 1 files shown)


modules/framework/src-jvm/RunnerCompat.scala
```diff
@@ -184,6 +187,13 @@ trait RunnerCompat[F[_]] { self: sbt.testing.Runner =>
       globalResources.traverse(_.sharedResources(map)).void
     }
 
+  private def spiGlobalResources: List[GlobalResourceF[F]] = {
+    val result = ListBuffer.empty[GlobalResourceF[F]]
+    val loader: ServiceLoader[GlobalResourceF[F]] = ServiceLoader.load(classOf[GlobalResourceF[F]])
```
Baccata, line 192: Hey @aevanszen, thanks for the PR. 

I gave it some long thought, and the problem with doing it this way is type-erasure. Even if it's unlikely that people would use different effect types in a single codebase, I'd be keen to protect against it. Because `GlobalResourceF` is parameterised, the `load` call would return everything, whether it's using cats-effect, monix or zio, and obviously loading monix globals in a CE context wouldn't be great. 

So what I think we should do instead, would be for the `ServiceLoader` call to be abstract away. We have a [SuiteLoader](https://github.com/disneystreaming/weaver-test/blob/c99cced2af84d7f5598bb1773d6e4c0c49f1d991/modules/framework/src/weaver/framework/SuiteLoader.scala) abstraction, and we could add a `def loadGlobalResources` to it. 

The implementation of `SuiteLoader` lives [there](https://github.com/disneystreaming/weaver-test/blob/3141fb7d8939d5adb350c205e9486db507fcdd70/modules/framework/src/weaver/framework/Fingerprints.scala#L37), at a place that is aware of the classes that the test-framework is supposed to understand. Those are effect-type specific (there is one set defined for CE, one for monix, one for zio etc). 

Now the thing is that you'll need one other layer of indirection to support cross-compilation. IE you could have a `PlatformServiceLoader` object defined in a `src-jvm`, containing a `spiGlobalResources[T](cl : ClassLoader, cls: Class[T]) : List[T]` method (similar to what you've implemented here), and a similar object defined in `src-js` , that'd have a method with the same signature but would return an empty list. 

This way, type-erasure would be accounted for.

Baccata:
I'd also appreciate : 

* a test (though I don't mind if we defer, unless you want to take a stab at it). 
* an amendment to the documentation page to describe the SPI-based mechanism

modules/framework/src-jvm/RunnerCompat.scala
aevanszen: Thank you for the detailed reply @Baccata, the suggestions are very helpful.

I had the same reservations around type-erasure but wasn't sure the best way to work around it in the framework just yet. Your suggestion is a good starting point, and I'll look into it.

I'll add the tests and documentation when the implementation is ok and update the PR with the changes.

The GlobalResoure classpath isn't an urgent blocker for me, so updating this PR may be slow as I work on it when I get the spare time. I will try to work on it, though, as it'll be useful to have.

Baccata: No problem whatsoever, I certainly appreciate the contribution, even if it takes time. There is no rush :)  let us know if you get stuck or if you want us to have a look at some commits

PR #518, opened 2022-05-03
Fix ZIOSuite test timings when clock is overridden

ollyw:
@Baccata as discussed a couple of weeks ago, here is a PR to fix the test timings when using a mocked clock in the shared layer.  This is my first weaver, so keen if you have any feedback

### DIFF
--- a/modules/core/zio/src/weaver/ziocompat/Live.scala
+++ b/modules/core/zio/src/weaver/ziocompat/Live.scala
@@ -1,0 +1,24 @@
+package weaver
+package ziocompat
+
+import zio.{ IO, URLayer, ZEnv, ZIO }
+
+/**
+ * Service used for getting the real environment during tests. This is useful
+ * for timing or getting random values during tests. For example, getting a
+ * random port for starting servers or timing an action with the real clock.
+ * This pattern is inspired by ZIO-test
+ */
+object Live {
+  trait Service {
+    def live[E, A](zio: ZIO[ZEnv, E, A]): IO[E, A]
+  }
+
+  def live[E, A](zio: ZIO[ZEnv, E, A]): ZIO[Live, E, A] =
+    ZIO.serviceWith[Service](_.live(zio))
+
+  def apply(): URLayer[ZEnv, Live] = ZIO.environment[ZEnv].map(env =>
+    new Service {
+      override def live[E, A](zio: ZIO[ZEnv, E, A]): IO[E, A] = zio.provide(env)
+    }).toLayer
+}
--- a/modules/core/zio/src/weaver/ziocompat/suites.scala
+++ b/modules/core/zio/src/weaver/ziocompat/suites.scala
@@ -49,9 +49,10 @@ abstract class BaseMutableZIOSuite[Res <: Has[_]](implicit tag: Tag[Res])
       else {
         for {
           ref <- Stream.eval(FiberRef.make(Chain.empty[Log.Entry]))
-          testLayer: RLayer[ZEnv, LogModule with ZEnv] =
-            ZEnv.any ++ ZLayer.fromService[Clock.Service, LogModule.Service](
-              FiberRefLog(ref, _))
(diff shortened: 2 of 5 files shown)


modules/core/zio/src/weaver/ziocompat/Live.scala
```diff
@@ -0,0 +1,24 @@
...
+
+import zio.{ IO, URLayer, ZEnv, ZIO }
+
+/**
+ * Service used for getting the real environment during tests. This is useful
+ * for timing or getting random values during tests. For example, getting a
+ * random port for starting servers or timing an action with the real clock.
+ * This pattern is inspired by ZIO-test
+ */
+object Live {
```
Baccata, line 12: This seems like a pretty overloaded term in zio-land. Any alternative name you can think of ?

Baccata, line 12: This seems like a pretty overloaded term in zio-land. Any alternative name you can think of ?

modules/core/zio/src/weaver/ziocompat/suites.scala
```diff
@@ -49,9 +49,10 @@ abstract class BaseMutableZIOSuite[Res <: Has[_]](implicit tag: Tag[Res])
       else {
         for {
           ref <- Stream.eval(FiberRef.make(Chain.empty[Log.Entry]))
-          testLayer: RLayer[ZEnv, LogModule with ZEnv] =
-            ZEnv.any ++ ZLayer.fromService[Clock.Service, LogModule.Service](
-              FiberRefLog(ref, _))
+          testLayer: RLayer[ZEnv, Live with LogModule with ZEnv] =
+            ZEnv.any >+> Live() ++ ZLayer.fromService[
```
Baccata, line 53: So basically if the user override any of the default `ZEnv` services in their sharedLayer, `Live` gives them a way to retain access to the `real` one, and if they don't, they just get the real layers ? Is my understanding correct ?

Baccata, line 53: So basically if the user override any of the default `ZEnv` services in their sharedLayer, `Live` gives them a way to retain access to the `real` one, and if they don't, they just get the real layers ? Is my understanding correct ?

modules/core/zio/src/weaver/ziocompat/Live.scala
ollyw: You are right that Live is an overloaded term, but AFAIK there is only one Live type/service in the zio libraries, and that is in zio-test for test purposes. It is familiar to zio-test users, who represent the majority of zio users. Only alternatives I can think of are `UnderlyingEnvironment` or `RealZEnv` or something similar. I am not sure these are better, but I'll carry on thinking

ollyw: You are right that Live is an overloaded term, but AFAIK there is only one Live type/service in the zio libraries, and that is in zio-test for test purposes. It is familiar to zio-test users, who represent the majority of zio users. Only alternatives I can think of are `UnderlyingEnvironment` or `RealZEnv` or something similar. I am not sure these are better, but I'll carry on thinking

modules/core/zio/src/weaver/ziocompat/suites.scala
ollyw: yes, exactly

ollyw: yes, exactly

modules/core/zio/src/weaver/ziocompat/Live.scala
Baccata: Cool, I'll trust your educated opinion on this :)

ollyw:
Sorry, i had a formatting error - I ran 'scalafmt' instead of 'scalafmtAll' and 'scalafix' instead of 'scalafixAll' :facepalm:  I have fixed, but the CI needs kicking off again

Baccata (approved):

PR #527, opened 2022-06-23
Improve determinism/coverage in Scalacheck tests

ollyw:
This PR changes the behaviour of the property test runner. It is motived by a strange situation in which a property test suite extending Checkers would fail intermittently. Superficially it seems that providing the same seed in the CheckConfig for the suite would provide the same test execution and be deterministic, however it turned out that is not the case.

The cause was that the failing property test generates seeds in a stream then runs the generators and the test _out of order_. Given the tests run until a threshold of minimumSuccessful has been reached, this means that if any tests happen to take longer, they do not necessarily complete before other faster tests have executed and reached the minimumSuccessful  threshold.

Here are two reasons why some tests are slower than others:
1) The tests that run first take longer. In our tests, the first 10 tests took around 50ms and subsequent ones took below 1ms. This could be to do with caches/JIT/gc/etc from test setup or something else. This happened regardless of the seed
2) Some seeds create tests that run longer. For example a test uses generators of list via Gen.listOf will generate lists between 0 and 100 by default, with a weighting of list length including generators with values 0, 1, 100 and 1-100. The max size of 100 is simply the default maximumGeneratorSize, but could be higher. It is entirely possible that a test that generates a list of length 100 will never complete before the tests with small list sizes satisfy the minimumSuccessful. This is "unfair" and not reassuring when the framework is effectively ignoring "difficult" tests in favour of "easy" ones completing. It might hide issues with algorithms with suboptimal algorithms (e.g. `O(n!)`). It would be better if the test explicitly times out.

There is a final frustration of the way that the test framework reports errors, in that the attempt that the test failed on varies between runs of the same seed, because there is no determinism in the reporting. e.g. "failed on try 10", "failed on try 15", but with the same seed causing the test to fail.

The proposal is to change the ordering of the tests within a property test stream o be _in order_. This restricts parallelism a little more then using an ordered stream. The flip side is that fewer redundant generations are discarded, so using less CPU overall. 

With this change, the expectation was that running a single property test within a suite _may_ execute slower, especially when the parallelism is restricted to 2-3 in `perPropertyParallelism`, but that the overall effect on the test suite would be that it would not take any longer, as the efficiency saving per test would work well where the number of suites/property tests were running in parallel.

In our example test module, with multiple executions, using the same seeds, here are the outputs from the bash `time` command, when run 5 times for unordered and ordered property tests:

```
Weaver 0.7.12 - baseline
mill -i -j 10 domain.test  146.91s user 3.69s system 1485% cpu 10.137 total
mill -i -j 10 domain.test  149.03s user 3.93s system 1507% cpu 10.147 total
mill -i -j 10 domain.test  145.58s user 3.95s system 1475% cpu 10.138 total
mill -i -j 10 domain.test  147.02s user 4.83s system 1519% cpu 9.995 total
mill -i -j 10 domain.test  149.30s user 4.27s system 1527% cpu 10.052 total

0.7.12 with the commit from this PR cherry-picked (to keep everything else constant)

mill -i -j 10 domain.test  145.79s user 4.50s system 1379% cpu 10.892 total
mill -i -j 10 domain.test  132.08s user 4.47s system 1402% cpu 9.734 total
mill -i -j 10 domain.test  135.11s user 4.64s system 1396% cpu 10.004 total
mill -i -j 10 domain.test  139.38s user 4.38s system 1421% cpu 10.111 total
mill -i -j 10 domain.test  136.49s user 4.46s system 1389% cpu 10.145 total
```

What can be seen is that the total CPU used with the ordered list is less, whilst the real world execution time hasn't change in any statistically significant way. More permutations were run and the results similar. That is not to say that for a codebase with highly variable test times between seeds, there might not be a slight change,. Perhaps in those cases, they will benefit from actually letting those slower tests complete, like we did on our project.

The one downside of this test is that unit testing can be a little hard. It is easy to create an example for manual testing by introducing a Thread.sleep in cases where the test fails, and prove that the other passing tests pass the whole property test before the failing test has had time to fail. However, this is not a good unit test, as tests with sleeps either have to sleep so long as to slow down the test (e.g. 10secs), or intermittently fail on sleep of 2 sec, depending if something else is using the CPU (e.g. GC, slack). I

If you can spot a good way to introduce a working test that is worth the effort, and will work on machines with a single vCPU, all the way up to 32+ cpus, that would be welcome. Otherwise, perhaps a test is not needed, given there are already tests for correctness, but not concurrency AFAIK.


Baccata:
@ollyw wow that's great, thanks so much for looking into it. I'm happy to accept this change as getting deterministic failures is important for property based testing. 

Regarding testing, here's some potential (not well caffeinated) ideas : 

I think using the [TestControl](https://typelevel.org/cats-effect/docs/core/test-runtime) construct would be a good way to unit-test this. You'd be constructing an IO[Expectations] using a forall, running it through the test runtime. This would prevent unit tests from taking long, even if they contain long sleeps. 

To get an input you could control, you could potentially create a contextual `Gen` instance that would perform a side effect of decrementing an atomic counter to get their values (10, 9, 8, ...) (yes, this is cheating). Then you would use the integers to perform an `IO.sleep`, which would mean that the values occurring first in the stream would take longer than the values occurring after. 

Then you introduce an arbitrary failure. Because the properties are supposed to be tested in parallel, the ones coming after would fail BEFORE (in terms of time) the ones coming before. Yet, because the fact that the properties are kept in order in the stream, it'd mean that you should still see the result of the property taking longer, every single time.

ollyw:
> @ollyw wow that's great, thanks so much for looking into it. I'm happy to accept this change as getting deterministic failures is important for property based testing.
> 
> Regarding testing, here's some potential (not well caffeinated) ideas :
> 
> I think using the [TestControl](https://typelevel.org/cats-effect/docs/core/test-runtime) construct would be a good way to unit-test this. You'd be constructing an IO[Expectations] using a forall, running it through the test runtime. This would prevent unit tests from taking long, even if they contain long sleeps.
> 
> To get an input you could control, you could potentially create a contextual `Gen` instance that would perform a side effect of decrementing an atomic counter to get their values (10, 9, 8, ...) (yes, this is cheating). Then you would use the integers to perform an `IO.sleep`, which would mean that the values occurring first in the stream would take longer than the values occurring after.
> 
> Then you introduce an arbitrary failure. Because the properties are supposed to be tested in parallel, the ones coming after would fail BEFORE (in terms of time) the ones coming before. Yet, because the fact that the properties are kept in order in the stream, it'd mean that you should still see the result of the property taking longer, every single time.

This suggestion seems good, although it looks like TestControl is only available in CE3. This means that the scalacheck project would have to either:
1) Have this test available only for CE3, by having version dependent sources
2) only have the test scope for CE3, probably via filtering the project matrix.
3) Not add this test anywhere

Do you have a preference @Baccata ?

Baccata:
Let's go with 1), the build already supports having CE-version-dependent sources. See example [here](https://github.com/disneystreaming/weaver-test/tree/master/modules/core)./

modules/scalacheck/test/src-ce3/weaver/scalacheck/CheckersConcurrencyTest.scala
```diff
@@ -0,0 +1,98 @@
...
+
+    val minTests = 10
+    object CheckersConcurrencyTestNested extends SimpleIOSuite with Checkers {
+
+      override def checkConfig: CheckConfig =
+        super.checkConfig.copy(perPropertyParallelism = minTests * 5,
+                               minimumSuccessful = minTests,
+                               maximumDiscardRatio = 10)
+
+      private val atomicInt = new AtomicInteger(0)
```
Baccata, line 29: I'd use a `Ref` in the `loggedTest` closure. Or at the very least move this atomicInt within the closure (as opposed for it being a field in a static object)

modules/scalacheck/test/src-ce3/weaver/scalacheck/CheckersConcurrencyTest.scala
```diff
@@ -0,0 +1,98 @@
...
+  test("tests should wait for slower tests to fail before completion") {
+
+    object CheckersConcurrencyTestNested extends SimpleIOSuite with Checkers {
+
+      override def checkConfig: CheckConfig =
+        super.checkConfig.copy(perPropertyParallelism = 50,
+                               minimumSuccessful = 10,
+                               maximumDiscardRatio = 10)
+
+      private val atomicInt = new AtomicInteger(0)
```
Baccata, line 68: ditto

Baccata:
> One particular area I was wondering about was the use of DogFood for runSuite, but it seemed to make things a little more complex, so I just implemented a simple solution. 

That's fair enough. DogFood suites are specifically useful for testing how the actual framework implementation behaves (with regards to the sbt-test-interface), it is un-necessary here.

modules/scalacheck/test/src-ce3/weaver/scalacheck/CheckersConcurrencyTest.scala
```diff
@@ -0,0 +1,98 @@
...
+      .flatMap { control =>
+        control.advanceAndTick(1.second).replicateA(minTests) *> control.results
+      }
+      .map {
+        case Some(Succeeded(r)) =>
+          expect(r.count(_.status == TestStatus.Success) == 1) and
+            expect(r.forall(_.status == TestStatus.Success)) and
+            expect.same(
+              expectedLogs,
+              expectedLogs.intersect(r.flatMap(_.log.toList.map(_.msg)))
```
Baccata, line 51: not sure about this  `intersect` call there.

Baccata:
Overall happy with how it's panning out 👍

modules/scalacheck/test/src-ce3/weaver/scalacheck/CheckersConcurrencyTest.scala
ollyw: My first reaction was to use a ref, but how would you execute it? Calling runUnsafe within the Gen.delay(Gen.const(...))  would be messy and is not really functional in the sense that the Generators are still side-effecting. If you can see another way, I am keen. However, using a functional construct, Ref, only to call it in a non-functional way seems messier then just using a clearly side-effecting construct.

modules/scalacheck/test/src-ce3/weaver/scalacheck/CheckersConcurrencyTest.scala
ollyw: because of parallelism there are additional runs of the generator (e.g. "Ran gen 11", "Ran gen 12"). These runs themselves are not a problem because the parallelism is speculative and works ahead. Using intersect was an easy way of filtering out the unneeded executions for the expecation

modules/scalacheck/test/src-ce3/weaver/scalacheck/CheckersConcurrencyTest.scala
Baccata: hah yeah, brainfart on my side. Let's move the atomicInt as part of the `loggedTest` block. Having a mutable var inside an `object` makes me uneasy 😅 , even if the object in question is nested under a lambda.

ollyw: Moving the AtomicInteger into the closure was a good idea :+1:  I have committed that change

modules/scalacheck/test/src-ce3/weaver/scalacheck/CheckersConcurrencyTest.scala
Baccata: Gotcha 👍

Baccata:
Great work, I'll wait for CI to be green and will merge/release. Thank you so much for taking this on, this is a very valuable change !

Baccata (approved):

PR #530, opened 2022-07-04
framework/cats: fix compilation warning regarding Dead Code (throw Boom)

bwiercinski:
no issue pinned

### DIFF
--- a/modules/framework/cats/test/src/DogFoodTests.scala
+++ b/modules/framework/cats/test/src/DogFoodTests.scala
@@ -130,7 +130,7 @@ object DogFoodTests extends IOSuite {
 
         // HONESTLY.
         val (location, capturedExpression) =
-          if (Platform.isScala3) (27, "1 == 2") else (28, "expect(1 == 2)")
+          if (Platform.isScala3) (31, "1 == 2") else (32, "expect(1 == 2)")
 
         val expected = s"""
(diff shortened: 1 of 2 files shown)


Baccata (approved):

Baccata:
Yeah, it's because the DogFoodTests aim at verifying the behaviour of the test-framework itself. In particular, it aims at verifying the fact that the macros capturing source files information (including line location) behave correctly. 

You've added lines, so the line number at which the assertions are written have changed, which means that the reported information has changed. You need to change the expected line numbers. See [here](https://github.com/disneystreaming/weaver-test/blob/master/modules/framework/cats/test/src/DogFoodTests.scala#L133). 

You may think "what the actual f*** ?" ... well, you're not wrong, but we're testing a test framework using that very test framework, so some things need to be hardcoded.

bwiercinski:
no no, its truly make sense :) fixed pipeline

modules/framework/cats/test/src/DogFoodTests.scala
```diff
@@ -130,7 +130,7 @@ object DogFoodTests extends IOSuite {
 
         // HONESTLY.
```
keynmol, line 131: This comment has never been truer.

keynmol (approved):

Issue #481, opened 2022-02-17
The Great Schism 


keynmol:
We've toyed with the idea of dropping 2.12 and CE2 support at some point.
I propose we roll upgrade to Scala 3.1 in this as well.

Roughly, here is the way I see it:

- [ ] `series/0.6.x` branch is created, it's frozen at CE2 and for bugfixes only
- [ ] CE2 modules are removed from main branch
- [ ] Monix modules are removed from main branch (unless Monix makes it onto CE3)
- [ ] CE-shims are removed and replaced with concrete CE3 typeclasses and std classes
- [x] Scala 3 version is bumped to 3.1

   ~~If this happens late enough, we might have access to `-scala-release` flag, and can generate 3.0-compatible Tasty code.~~
   ~~That said, with the adoption figures the way they are, I say we encourage upgrade to 3.1.~~

   TEAM UPGRADE REPRESENT YO #542 

- [ ] (?) Semantics are adjusted to better take advantage of CE3 improvements
- [ ] 0.8.0 is released with only CE3 support
- [ ] Mima is enabled within the 0.8.x lineage (according to early-semver)

What do you think, @Baccata ?

Baccata:
Yup, that's pretty much what I was thinking, so happy for this plan to be actioned!

nlchar:
Will Scala 2 be still supported?

keynmol:
Yes, I don't envision Scala 2 support going away at least in the couple of years.

nlchar:
Thanks. Good to hear since almost all the community is still on Scala 2.

keynmol:
All but one of those have been enacted, so I'm closing this with the impending release of a slimmed down 0.8.0

PR #542, opened 2022-07-06
Upgrade libraries, Scala.js and Scala 3

### LINKED ISSUE #476: Make a decision with Scala.js macrotask executor
I've actively avoided reading too much about the impact of it, but we need to make a decision explicit to be able to keep up with SJS upgrades:

```
[error] /home/runner/work/weaver-test/weaver-test/modules/core/cats/src-ce2/weaver/CatsUnsafeRun.scala:14:38: The global execution context in Scala.js is based on JS Promises (microtasks).
[error] Using it may prevent macrotasks (I/O, timers, UI rendering) from running reliably.
[error] 
[error] Unfortunately, there is no way with ECMAScript only to implement a performant
[error] macrotask execution context (and hence Scala.js core does not contain one).
[error] 
[error] We recommend you use: https://github.com/scala-js/scala-js-macrotask-executor
[error] Please refer to the README.md of that project for more details regarding
[error] microtask vs. macrotask execution contexts.
[error] 
[error] If you do not care about macrotask fairness, you can silence this warning by:
[error] - Adding @nowarn("cat=other") (Scala >= 2.13.x only)
[error] - Setting the -P:scalajs:nowarnGlobalExecutionContext compiler option
[error] - Using scala.scalajs.concurrent.JSExecutionContext.queue
[error]   (the implementation of ExecutionContext.global in Scala.js) directly.
[error] 
[error] If you do not care about performance, you can use
[error] scala.scalajs.concurrent.QueueExecutionContext.timeouts().
[error] It is based on setTimeout which makes it fair but slow (due to clamping).
[error]                 
[error]     IO.contextShift(ExecutionContext.global)
[error]                                      ^
```

https://github.com/disneystreaming/weaver-test/runs/5188229336?check_suite_focus=true

keynmol:
Also bring in the macrotask dependency for Scala.js

I don't think the decision to bump Scala to 3.1.x is controversial anymore - Cats did it in 2.8.0, 20 days ago.

This helps clear out the path of upgrades to the great schism (#481), giving us a sort of cleanish slate.

I couldn't do `zio 2.5.1.0 -> 2.5.1.1` because I got 59 compilation errors and... yeah.

If this goes green, I'll just close all the individual steward PRs and iterate here instead.

Closes #476

### DIFF
--- a/.github/workflows/ci.yml
+++ b/.github/workflows/ci.yml
@@ -15,7 +15,7 @@ jobs:
       matrix:
         os: [ubuntu-latest]
         java: [adopt@1.8]
-        scalaVersion: ["2_12", "2_13", "3_0"]
+        scalaVersion: ["2_12", "2_13", "3"]
         scalaPlatform: ["jvm", "js"]
         ceVersion: ["CE2", "CE3"]
     runs-on: ${{ matrix.os }}
--- a/.scalafmt.conf
+++ b/.scalafmt.conf
@@ -1,4 +1,4 @@
-version = "3.5.3"
+version = "3.5.8"
 
 runner.dialect=scala213source3
 
--- a/build.sbt
+++ b/build.sbt
@@ -36,7 +36,7 @@ Global / (Test / testOptions) += Tests.Argument("--quickstart")
 
 val Version = new {
   object CE3 {
-    val fs2        = "3.1.6"
+    val fs2        = "3.2.9"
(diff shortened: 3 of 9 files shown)


Baccata:
Thanks a LOT for this, after the wave of upgrades, I figured I'd have to add it to my ever growing TODO list. Really glad I don't have to.

Baccata (approved):

PR #543, opened 2022-07-06
The Great Schism: Remove Cats Effect 2 and monix

keynmol:
This PR intends to do everything in #481 

TODO: 
- [x] Remove CECompat
- [ ] Remove any modules we want to stop publishing?


modules/core/src/weaver/Runner.scala
```diff
@@ -2,27 +2,34 @@ package weaver
...
   import Runner._
 
   // Signaling option, because we need to detect completion
-  private type Channel[A] = CECompat.Queue[F, Option[A]]
+  private type Channel[A] = Queue[F, Option[A]]
+
+  private[this] def background[A, B](fa: F[A], default: A)(
+      f: F[A] => F[B]): F[B] =
+    Async[F].background(fa).use(fOutcome =>
+      f(fOutcome.flatMap(_.embed(onCancel = Async[F].pure(default)))))
```
keynmol, line 25: Inline this, I thought it was used in more places.

Issue #528, opened 2022-06-29
Use both Discipline and MutableIOSuite


bwiercinski:
is it possible to use **shared resource** from MutableIOSuite inside **checkAll** from Discipline addon?

Baccata:
Mmm yeah, that's a fair point I guess. In this case I'd be happy to review a PR adding this feature 😄

bwiercinski:
grrr 😄 

how to achive this? currently `Discipline` trait have a self-type of `FunSuiteAux` (`trait Discipline { self: FunSuiteAux => ... }`)

obviously i have to use something that relays on `MutableFSuite` (which is a abstract class). should i create trait `Discipline2` which have `trait DisciplineIT[F[_]] { self: MutableFSuite[F] => ... }`?

if i would like to edit current Discipline trait it would be binary incompatible although the users would only need to change:
```diff
+object FooLawTests extends FunSuite with Discipline {
-object FooLawTests extends SimpleMutableIOSuite with Discipline {

  checkAll("foo", FooTests(implementation).foo)

+  pureTest("if someone added any unit test) { ... }
-  test("if someone added any unit test) { ... }
}
```

Baccata:
Yeah let's try to make a new trait instead (could call it `DisciplineEffect` or whatever), in order to avoid breaking current usecases. Not that I think there are many of the, but still.

bwiercinski:
I'm in a crossroad right now. At first I wanted to use `trait DisciplineEffect[F[_]] { self: MutableFSuite[F] => ... }`. Its not so easy:
In standard suite all tests are collected using `MutableFSuite.registerTest`. If after filtering number of tests is 0 then there is no need to allocate resource.

In case of law testing I want to define following function `checkAll(name: String, input: Res => F[Laws#RuleSet])` (one `Laws#RuleSet` contains many tests). So I want to firstly open a resource and then register tests, because without the resource I dont know how many tests there will be.

* I can modify `MutableFSuite` to contain instead of 
```diff
+var testSeq = Seq.empty[(TestName, Res => F[TestOutcome])]
-var testSeq = Seq.empty[Res => F[List[(TestName, TestOutcome)]]]
```
I'm losing benefit of not opening a resource in some cases. **Is that something to concern or hardly ever used feature?**

* I can modify `MutableFSuite` to contain instead of 
```diff
+var testSeq = Seq.empty[(TestName, Res => F[TestOutcome])]
[...]

PR #548, opened 2022-07-12
create DisciplineFSuite #528

bwiercinski:
https://github.com/disneystreaming/weaver-test/issues/528

### DIFF
--- a/modules/core/src-jvm/weaver/junit/WeaverRunner.scala
+++ b/modules/core/src-jvm/weaver/junit/WeaverRunner.scala
@@ -18,7 +18,7 @@ class WeaverRunner(cls: Class[_], dummy: Boolean)
     Reflection.loadRunnableSuite(cls.getName(), getClass().getClassLoader())
   }
 
-  lazy val testDescriptions: Map[String, Description] = {
+  def testDescriptions: Map[String, Description] = {
     suite.plan.map(name =>
       name.name -> Description.createTestDescription(cls, name.name)).toMap
   }
--- a/modules/discipline/src/weaver/discipline/Discipline.scala
+++ b/modules/discipline/src/weaver/discipline/Discipline.scala
@@ -14,6 +14,7 @@ import org.scalacheck.util.Pretty
 import org.typelevel.discipline.Laws
 import org.scalacheck.Prop
 import Discipline._
+import scala.collection.concurrent.TrieMap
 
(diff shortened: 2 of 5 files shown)


Baccata:
I think it's good. Let's see how it looks like in tests 😄

bwiercinski:
@Baccata can you check it out. i was managed to implement it using `RunnableSuite`. it works well both in sbt and in intellij, but with that price I had to reimplement `WeaverRunner`. now `testDescriptions` is evaluated each time instead of lazy val. are you ok with that?

if it is ok I will fix JS tests. or maybe you have more clever idea

Baccata:
@bwiercinski , mmm, I'm gonna have to give it a think. I'll get back to you next week

bwiercinski:
did you run it? is reevaluating testDescriptions each time a big problem? maybe you know better edgecases using junit api with intellij

Baccata:
> maybe you know better edgecases using junit api with intellij

I'm actually not certain ... did you try to run a DisciplineF spec with Intellij ? If you could provide a screen capture to see what it looks like, that'd be great, with several tests/rule sets being tested.  (assuming you're an Intellij user, which I'm not, and I've uninstalled Intellij a while ago, would rather not have to download it again).

bwiercinski:
as they are evaluated they appearing on the GUI, firstly **1 of 1 tests passed**, then 2/2 and so on. they appear in the order as they are computed, but i can easily sort them by name by using a button. not best but looks good.
image

using sbt:
image

non discipline tests are working as usual
image

Baccata:
Okay, I **think** it's good enough for me. Proceed with some tests ^^

bwiercinski:
all passed :)

Baccata:
Thanks @bwiercinski, I'm not forgetting you, I've just been incredibly busy. I'll review later this week

Baccata (approved):

Baccata:
Thanks for this !

bwiercinski:
thanks :) , when you will release it?

Baccata:
On its way : https://github.com/disneystreaming/weaver-test/releases/tag/v0.6.14

bwiercinski:
i screwed up when using synchronized. monitors used on synchronized are `var`s and eventually synchronization doesn't work. it only matters when using intellij. i will make a new PR with a fix

PR #559, opened 2022-09-02
Ignored test name tag

callado4:
Add an `ignored` test name tag to easily ignore individual tests.

We had some internal discussions about adding something like this.

The only thing I'm not sure of is using `.ignored` vs `.ignore` (including the "d" or not)

### DIFF
--- a/modules/core/src/weaver/Expectations.scala
+++ b/modules/core/src/weaver/Expectations.scala
@@ -150,8 +150,8 @@ object Expectations {
     }
 
     implicit class StringOps(str: String) {
-      def ignored(implicit loc: SourceLocation): TestName =
-        new TestName(str, loc, Set.empty).ignored
+      def ignore(implicit loc: SourceLocation): TestName =
+        new TestName(str, loc, Set.empty).ignore
       def only(implicit loc: SourceLocation): TestName =
         new TestName(str, loc, Set.empty).only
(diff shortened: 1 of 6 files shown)


keynmol:
I'm not sure whether this adds anything to the already existing ability to ignore the tests: https://disneystreaming.github.io/weaver-test/docs/tagging ? 

The docs, I now realise, are not setup in a fantastic fashion, but here's what it looks like: 
image

kubukoz:
@keynmol existing ignore doesn't work in pure tests.

modules/core/src/weaver/Expectations.scala
```diff
@@ -150,6 +150,8 @@ object Expectations {
     }
 
     implicit class StringOps(str: String) {
+      def ignore(implicit loc: SourceLocation): TestName =
```
kubukoz, line 153: IMO `ignored` is more consistent with `tagged`

kubukoz (changes requested):

modules/core/src/weaver/Expectations.scala
Baccata: We'll keep `ignore` on the account that it's consistent with Munit : https://scalameta.org/munit/docs/filtering.html#ignore-single-test-case

Baccata (approved):

Baccata:
Thanks !

PR #568, opened 2022-09-13
Scala-native support

Baccata:
Adds Scala-Native support 

Waiting for this to be merged and for the release train to start (we need CE + fs2 non-RC versions) : https://github.com/typelevel/cats-effect/pull/3138


armanbilge:
Nice! Is there a possibility of a snapshot, or should I bootleg this? Looking to unblock the fs2-data test suite.

Baccata:
@armanbilge if you want it today you're gonna need to do a bootleg, but we'll set up snapshot publishing before the end of the week for the series/0.8 branch.

build.sbt
```diff
@@ -38,12 +38,12 @@ sonatypeCredentialHost := "s01.oss.sonatype.org"
 
 val Version = new {
   object CE3 {
-    val fs2        = "3.2.12"
-    val cats       = "3.3.14"
+    val fs2        = "3.2.14-75-7902cbf"
+    val cats       = "3.3.14-1-5d11fe9"
```
keynmol, line 42:
```suggestion
    val cats       = "3.3.14"
```

https://github.com/typelevel/cats-effect/issues/1302#issuecomment-1247387034

modules/core/src-native/weaver/internals/Timestamp.scala
```diff
@@ -0,0 +1,35 @@
...
+import scala.scalanative.posix
+import scala.scalanative.unsafe._
+
+import posix.time
+import posix.timeOps._
+
+private[weaver] object Timestamp {
+
+  def format(epochSecond: Long): String = Zone { implicit zone =>
+    val out     = alloc[time.tm]()
```
keynmol, line 12: I think this can be `stackalloc`, no? 

We're not return the pointer to the caller, so we don't need it on the heap, and the return value is a Scala String which will be on the heap anyways.

modules/core/src-native/weaver/internals/Timestamp.scala
```diff
@@ -0,0 +1,35 @@
...
+    val gmTime: Ptr[time.tm] = time.localtime_r(timePtr, out)
+    val hour                 = gmTime.tm_hour
+    val minutes              = gmTime.tm_min
+    val seconds              = gmTime.tm_sec
+    s"$hour:$minutes:$seconds"
+  }
+
+  def localTime(hours: Int, minutes: Int, seconds: Int): Long = Zone {
+    implicit zone =>
+      val out     = alloc[time.tm]()
```
keynmol, line 24: same as above - the return is `Long`, by value, so the memory can be allocated by `stackalloc` - this would make it much, much faster as well

modules/core/src-native/weaver/internals/Timestamp.scala
keynmol: And this makes Zone unnecessary as well

modules/framework/src-js/PlatformTask.scala
```diff
@@ -0,0 +1,19 @@
...
+
+private[weaver] trait PlatformTask extends AsyncTask {
+
+  override def execute(
+      eventHandler: EventHandler,
+      loggers: Array[Logger],
+      continuation: Array[Task] => Unit): Unit = {
+    val _ = executeFuture(eventHandler, loggers).map(_ =>
+      continuation(Array.empty[Task]))(
+      scala.scalajs.concurrent.JSExecutionContext.queue)
```
keynmol, line 13: Hmm, is this what we had before?
IIRC we are bringing the macrotask - do we want to use it instead? https://github.com/disneystreaming/weaver-test/issues/476

modules/framework/src/weaver/framework/DogFood.scala
```diff
@@ -28,8 +28,8 @@ abstract class DogFood[F[_]](
   // for some time before getting the logs back. On JVM platform
   // we do not need to wait, since the suite will run synchronously
   private val patience: Option[FiniteDuration] = PlatformCompat.platform match {
-    case JS  => 2.seconds.some
-    case JVM => none
+    case JS | Native => 2.seconds.some
```
keynmol, line 31: For my own understanding - the comment above talks about SJS being asynchronous, hence the need for waiting.

Do we still need it for Native, which will also execute synchronously? Or is it something else?

modules/core/src-native/weaver/internals/Timestamp.scala
Baccata: yeah fair enough, I copied it verbatim from 0.4.0 code I had laying around

modules/framework/src-js/PlatformTask.scala
Baccata: In this case it really doesn't matter

Baccata: > Hmm, is this what we had before?

Nah before we were using a callback based async method. We're unifying between JS and Native by relying on Future instead, and we're only needing an EC to plug the "end of the world" callback to the future.

modules/framework/src/weaver/framework/DogFood.scala
Baccata: [aa7296e](https://github.com/disneystreaming/weaver-test/pull/568/commits/aa7296e2be5a9ba2cd6ccad7cda27c5b76c76b81)

modules/core/src-native/weaver/internals/Timestamp.scala
Baccata: Addressed [aa7296e](https://github.com/disneystreaming/weaver-test/pull/568/commits/aa7296e2be5a9ba2cd6ccad7cda27c5b76c76b81)

modules/core/src-native/weaver/internals/Timestamp.scala
Baccata: Addressed [aa7296e](https://github.com/disneystreaming/weaver-test/pull/568/commits/aa7296e2be5a9ba2cd6ccad7cda27c5b76c76b81)

build.sbt
```diff
@@ -38,12 +38,12 @@ sonatypeCredentialHost := "s01.oss.sonatype.org"
 
 val Version = new {
   object CE3 {
-    val fs2        = "3.2.12"
+    val fs2        = "3.2.14-75-7902cbf"
```
keynmol, line 41:
```suggestion
    val fs2        = "3.3.0"
```

keynmol (approved):

Baccata:
I'm just gonna go ahead and merge this, we can PR the fs2 version later on against the 0.8 branch

PR #571, opened 2022-09-14
Make parallel/sequential suites in DogFood lazy resources tests deterministic

### LINKED ISSUE #565: Make dogfood tests deterministic
Currently some dog-food tests are utterly underterministic. They could be made so by using [cats-effect-testkit](https://typelevel.org/cats-effect/docs/core/test-runtime) 

Example of failed run : https://github.com/disneystreaming/weaver-test/actions/runs/3035811871/jobs/4886387716

keynmol:
1. for LazyAccessParallel tests, we use CyclicBarrier to ensure that all 3 suites enter the assertions block at the same time - this will keep the lazy resource alive for long enough to be reused, hence satisfyingthe test
2. For LazyAccessSequential, we use a semaphore with 1 permit to make sure the tests are indeed sequentially executed and finalised before the next one starts. This helps testing that lazy resource is indeed reinstantiated on demand if the suite that originally created it has already finished

Closes #565

### DIFF
--- a/modules/framework/cats/test/src-jvm/MetaJVM.scala
+++ b/modules/framework/cats/test/src-jvm/MetaJVM.scala
@@ -4,10 +4,8 @@ package test
 
 import java.io.File
 
-import scala.concurrent.duration._
-
 import cats.effect._
-import cats.effect.std.CyclicBarrier
+import cats.effect.std.{ CyclicBarrier, Semaphore }
 import cats.syntax.all._
 
 // The build tool will only detect and run top-level test suites. We can however nest objects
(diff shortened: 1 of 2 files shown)


modules/framework/cats/test/src-jvm/MetaJVM.scala
```diff
@@ -64,14 +69,24 @@ object MetaJVM {
     def sharedResources(global: weaver.GlobalWrite): Resource[IO, Unit] =
       Resource.eval {
         for {
-          initialised <- Ref[IO].of(0)
-          finalised   <- Ref[IO].of(0)
-          totalUses   <- Ref[IO].of(0)
+          initialised   <- Ref[IO].of(0)
+          finalised     <- Ref[IO].of(0)
+          totalUses     <- Ref[IO].of(0)
+          parallelLatch <- CyclicBarrier[IO](3)
```
keynmol, line 75: The thinking here is that the suites could use this latch to make sure that all of them are running at the same time

Once the barrier is reached, they can run the assertions on the state.

The lazy resource will stay open and be re-used, because no tests will finish.

Using a fixed number here is problematic because it hardcodes the physical number of tests running in parallel - if we add another test to `LazyAccessParallel` spec, it will break again.

Baccata: > Using a fixed number here is problematic because it hardcodes the physical number of tests running in parallel 

This is fine, let's just put a comment above it.

keynmol: done

Baccata (approved):

Issue #590, opened 2022-10-05
Fail tests marked with `.only` if running in CI


kubukoz:
Similarly to what munit does - if the `CI` environment flag is set, Weaver should fail any test marked with the `.only` tag.

kubukoz:
duped by #591

kubukoz:
OR IS IT

PR #598, opened 2022-11-01
Fail tests marked with .only if running in CI

callado4:
Implements https://github.com/disneystreaming/weaver-test/issues/590

### DIFF
--- a/modules/core/src/weaver/suites.scala
+++ b/modules/core/src/weaver/suites.scala
@@ -110,13 +110,14 @@ abstract class MutableFSuite[F[_]] extends RunnableSuite[F]  {
         }
       } else onlyTestsNotIgnored.map(_._2)
       val parallism = math.max(1, maxParallelism)
-      if (filteredTests.isEmpty) Stream.empty // no need to allocate resources
-      else if (testsTaggedOnly.nonEmpty && isCI) {
+
+      if (testsTaggedOnly.nonEmpty && isCI) {
         val failureOutcomes = testsTaggedOnly
           .map(_._1)
           .map(onlyNotOnCiFailure)
         Stream.emits(failureOutcomes).lift[F](effectCompat.effect)
       }
+      else if (filteredTests.isEmpty) Stream.empty // no need to allocate resources
       else for {
         resource <- Stream.resource(sharedResource)
         tests = filteredTests.map(_.apply(resource))
--- a/modules/core/src/weaver/suites.scala
+++ b/modules/core/src/weaver/suites.scala
@@ -1,5 +1,8 @@
 package weaver
 
+import scala.concurrent.duration.FiniteDuration
+
+import cats.data.Chain
 import cats.effect.{ Async, Resource }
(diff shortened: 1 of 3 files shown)


callado4:
Something seems wrong since github believes my fork's master is in sync based on commit 44c2ba but master currently should be much later.  I need to do some git-fu and update my PR (I want it to be based off the master in this repo)

modules/core/src/weaver/suites.scala
```diff
@@ -114,6 +126,20 @@ abstract class MutableFSuite[F[_]] extends RunnableSuite[F]  {
...
+      source = None,
+      location = List(test.location)
+    )
+    TestOutcome(
+      name = test.name,
+      duration = FiniteDuration(0, "ns"),
+      result = result,
+      log = Chain.empty
+    )
+  }
```
callado4, line 141: I think this makes sense for making a failing test outcome, but if there is a better way, please let me know.

A few things that I wanted to clarify
- msg in the result is the same message used in Munit
- source in result is None, this was't caused by an exception so no need to provide an exception (right?)
- location in result, using the location of the test definition itself as a single entry, is this what is expected here?
- duration is TestOutcome set to 0 since this test didn't actually run - is this fine?
- I provided an empty chain for TestOutcome.log, but do we want something here?

modules/core/src/weaver/suites.scala
```diff
@@ -92,19 +95,28 @@ abstract class MutableFSuite[F[_]] extends RunnableSuite[F]  {
...
-      } else testsTaggedOnly.map(_._2)
+      } else onlyTestsNotIgnored.map(_._2)
       val parallism = math.max(1, maxParallelism)
       if (filteredTests.isEmpty) Stream.empty // no need to allocate resources
+      else if (testsTaggedOnly.nonEmpty && isCI) {
+        val failureOutcomes = testsTaggedOnly
+          .map(_._1)
+          .map(onlyNotOnCiFailure)
+        Stream.emits(failureOutcomes).lift[F](effectCompat.effect)
+      }
```
callado4, line 119: Do we want the tests to fail if there would end up being no tests to run, but there was something tagged `.only`? For example, a suite with just one test that has both `.only` and `.ignore` would be caught in the first if condition here and so we would return a `Stream.empty`, i.e., no tests would be run at all (and the `.only` is ignored).

On the other hand, if there are _any_ tests to be run, and any of the other tests are marked as `.only` and `.ignore` we will fail (we will report all tests marked with `.only`)

Baccata: I think I'd swap both conditions so that the `isCI` flag would be checked first, and result in a failure if enabled and in the presence of any `.only`. The reason is that the behaviour would be surprising in a later PR when tests are added and there might be a cognitive disconnect between the suite failing and the addition of `.only`.

callado4: done

Baccata (approved):

Baccata:
Thanks !

PR #606, opened 2022-11-15
Update versions

Baccata:


keynmol:
Just 1 comment, even though you know my position on upgrades - TEAM UPGRADE YO.

The only reason people _need_ to upgrade to CE3.4. is to re-enable Graalvm Native Image on Scala 3 - hence the bump to 3.2.x, as it needs entirely new bytecode.

We as testing framework don't _need_ to upgrade, but we still can. Thought I wonder if we should wait for a little bit for 3.4.0 to propagate first?

Baccata:
> We as testing framework don't need to upgrade, but we still can.

Asking ourselves that question contributes to the maintenance burden. We're not promising any LTS on any of the (non-LTS) Scala 3 minor releases, and I can't really be bothered with delaying upgrades out of concern for the handful of Scala 3 users who are on non-LTS versions (the future 3.3 being the first planned LTS, afaik). 

Do you have a non-hypothetical case to point to, that would be impacted by these bumps ? (my point of view might change if you do).

keynmol:
Nah, I wasn't really even trying to block this upgrade - just wanted to spell out the possible reasons not to upgrade.

I think with dispatcher deprecations in CE 3.4 this upgrade will fail the build if the users are using strict flags, but they can always rollback.

So all 🟢 on my side.

keynmol (approved):

Baccata:
> but they can always rollback. 

Yup that's my thinking too

Issue #614, opened 2022-11-27
Suggestion: partial pattern match syntax


mcanlas:
What do we think? Tried to provide more vibrant information by using `Show`

Inspired by https://www.scalatest.org/user_guide/using_inside

```scala
def matches[A](x: A)(
      f: PartialFunction[A, Expectations]
  )(implicit pos: SourceLocation, A: Show[A] = Show.fromToString[A]): Expectations =
    if (f.isDefinedAt(x))
      f(x)
    else
      failure("Pattern did not match, got: " + A.show(x))
```

mcanlas:
Tangent Q: thoughts on a matcher for something like `MonadError` and dump the error channel on failure?

Baccata:
> What do we think? 

I'm not certain I want it, not certain I don't want it. @keynmol wdyt ?

> Tangent Q: thoughts on a matcher for something like MonadError and dump the error channel on failure?

You can just lift your monad to the effect type and gain this behaviour.

mcanlas:
Re: the `MonadError` case, I have a suite that is pure and one of the return values happens to be `Either` or `ValidatedNec`. So it's inconvenient to hoist up the suite's effect type just for singular cases.

keynmol:
> I'm not certain I want it, not certain I don't want it. @keynmol wdyt ?

Funny that Mark mentioned `inside` specifically - I've been using it a _ton_ in my current scalatest codebase, and IMO it's an excellent combinator. I'm using it to replace a lot of the usages of `.get`/`.fold` on Eithers.

It mostly shows up in doing tons of this:

```scala
  enum Test:
    case Yo(i: Int) 
    case Noyo(b: String)

  exists(Option(Test.Yo(15))) {
    case Test.Yo(i) => expect(i == 25)
    case Test.Noyo(b) => failure("expected Yo")
  }.run.tap(println)
```

having to handle the "rest" branch and give it a meaningful message is annoying.
A generic case of this is asserting on the Left part of either if you do `attempt` or something.

Currently our forEach and exists combinators are not sufficient:

```scala
  val x = Try[Int](throw NoSuchMethodException("yo"))

  forEach(x) {i => expect(i == 25)}.run.tap(println)
  forEach(x.toEither) {i => expect(i == 25)}.run.tap(println)
  exists(x) {i => expect(i == 25)}.run.tap(println)
  exists(x.toEither) {i => expect(i == 25)}.run.tap(println)
```

output:

```
Valid(())
[...]

PR #615, opened 2022-11-29
RFC: matches and whenSuccess combinators

mcanlas:
Includes two independent proposals per #614

1) `matches` partial pattern match combinator
2) `existsRight` combinator for asserting `ApplicativeError`

Proposal 2 is more a strawman, I'm not confident in it at all. I had to also demand `Foldable`. I looked at `Bifoldable` briefly but I wasn't sure how to merge the two channels into one. Felt like that was describing two fully independent channels vs a disjoint union in an `ApplicativeError`

### DIFF
--- a/modules/core/src/weaver/Expectations.scala
+++ b/modules/core/src/weaver/Expectations.scala
@@ -217,7 +217,7 @@ object Expectations {
      *
      * @example
      *   {{{
-     *     matches(Option(4) { case Some(x) =>
+     *     matches(Option(4)) { case Some(x) =>
      *       expect.eql(4, x)
      *     }
      *   }}}
--- a/modules/core/src/weaver/Expectations.scala
+++ b/modules/core/src/weaver/Expectations.scala
@@ -238,6 +238,19 @@ object Expectations {
     def inEach[L[_], A](la: L[A])(f: A => Expectations)(
         implicit L: Foldable[L]): Expectations = forEach(la)(f)
 
+    /**
+     * Checks that an `ApplicativeError` (like `Either`) is successful
+     *
+     * @example
+     *   {{{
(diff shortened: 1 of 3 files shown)


modules/core/src/weaver/Expectations.scala
```diff
@@ -211,12 +211,45 @@ object Expectations {
...
+        f: A => Expectations
+    )(
+        implicit pos: SourceLocation,
+        F: ApplicativeError[F, E],
+        G: Foldable[F],
+        E: Show[E] = Show.fromToString[E]): Expectations =
+      fa
+        .map(f)
+        .handleError(e => failure(e.show))
+        .foldLeft(failure("impossible?"))((_, x) => x)
```
keynmol, line 251: May be `.attempt.map(_.fold(...))`?

mcanlas: Wasn't totally sure what the goal here would be. Doesn't something need to interpret `F`? In the PR's case, that's what `F : Foldable` is for

`attempt` would still be wrapped in `F`

mcanlas:
Renamed to `whenSuccess`

Wasn't sure about the `Foldable` logic. I realized the left `b` value I was ignoring initially was the accumulator value?

Tried to clarify the logical edge case error messages

What still needs doing (potentially after a thumbs up) is documenting these combinators

mcanlas:
Documentation added!

Baccata (approved):
Thanks ! I'll wait for @keynmol's feedback before merging.

keynmol (approved):

callado4:
Any idea when this will be officially released? I would like to use this functionality

Baccata:
Thanks for the reminder. I've triggered a release, it should be here today.

PR #633, opened 2023-02-06
Use builder pattern for CheckConfig

### LINKED ISSUE #632: CheckConfig should accept `org.scalacheck.rng.Seed` instead of `Long`
When a test fails it reports a base64 encoded `Seed`, from which we can attempt to construct it back with `org.scalacheck.rng.Seed#fromBase64`, but `CheckConfig` accepts `Long`.

dabd:
### What
* Implement `CheckConfig` using the builder pattern (tweaked the output of https://github.com/alexarchambault/data-class).
* This PR is a breaking change but hopefully the class can be extended in a backwards compatible way.

### Why
* Fix https://github.com/disneystreaming/weaver-test/issues/632.

### DIFF
--- a/modules/scalacheck/src/weaver/scalacheck/CheckConfig.scala
+++ b/modules/scalacheck/src/weaver/scalacheck/CheckConfig.scala
@@ -1,18 +1,106 @@
 package weaver
 package scalacheck
 
-case class CheckConfig(
-    minimumSuccessful: Int,
-    maximumDiscardRatio: Int,
-    maximumGeneratorSize: Int,
-    perPropertyParallelism: Int,
-    initialSeed: Option[Long]
-) {
+import org.scalacheck.rng.Seed
+
+final class CheckConfig(
+    val minimumSuccessful: Int,
+    val maximumDiscardRatio: Int,
+    val maximumGeneratorSize: Int,
+    val perPropertyParallelism: Int,
+    val initialSeed: Option[Seed]) extends Product with Serializable {
   assert(maximumDiscardRatio >= 0)
   assert(maximumDiscardRatio <= 100)
   assert(minimumSuccessful > 0)
 
   def maximumDiscarded = minimumSuccessful * maximumDiscardRatio / 100
+
+  def withMinimumSuccessful(minimumSuccessful: Int) = new CheckConfig(
+    minimumSuccessful = minimumSuccessful,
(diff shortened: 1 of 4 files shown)


modules/scalacheck/src/weaver/scalacheck/CheckConfig.scala
```diff
@@ -1,26 +1,70 @@
...
-    maximumGeneratorSize: Int,
-    perPropertyParallelism: Int,
-    initialSeed: Option[Long]
-) {
-  assert(maximumDiscardRatio >= 0)
-  assert(maximumDiscardRatio <= 100)
-  assert(minimumSuccessful > 0)
+import org.scalacheck.rng.Seed
+
+sealed abstract class CheckConfig private { outer =>
```
dabd, line 6: Using a similar pattern than in scalacheck for `Test.Parameters`.
https://github.com/typelevel/scalacheck/blob/main/core/shared/src/main/scala/org/scalacheck/Test.scala#L23

modules/scalacheck/src/weaver/scalacheck/CheckConfig.scala
```diff
@@ -1,26 +1,70 @@
...
+  val default: CheckConfig = new CheckConfig {
+    val minimumSuccessful      = 80
+    val maximumDiscardRatio    = 5
+    val maximumGeneratorSize   = 100
+    val perPropertyParallelism = 10
+    val initialSeed            = None
+  }
+
+  def apply(
+      minimumSuccessful0: Int,
```
dabd, line 55: Is there a way to avoid the ugly names (with suffix 0) here?

dabd:
If it is acceptable to have a [compile-time only dependency](https://stackoverflow.com/questions/21515325/add-a-compile-time-only-dependency-in-sbt/21516954#21516954) could use https://github.com/alexarchambault/data-class.

keuhdall:
Hi ! What's the status of this PR ? I'm experiencing the same issues when dealing with seeds atm ☹️

dabd:
> Hi ! What's the status of this PR ? I'm experiencing the same issues when dealing with seeds atm ☹️

Because this PR is flagged as binary breaking, I am not sure if/when it will be merged 🤷

dabd:
Resolved by https://github.com/typelevel/weaver-test/pull/39.

PR #648, opened 2023-04-15
Auto-derive Show and PropF instances

EstebanMarin:
# The purpose is to  Auto-derive Show and PropF instances

- Reduce user effort in creating instances
- Enhance library usability

# Changes

```
Import cats.derived.auto.show._:
```

This import brings the functionality to automatically derive instances of the Show typeclass for case classes and tuples. This helps to reduce boilerplate code, as you won't need to manually create instances for each of the tuples used in the tests.

Add the autoShowTuple implicit def:
This function automatically derives Show instances for tuples of two elements (A, B) where both A and B have their own Show instances. This further simplifies the code, as you won't need to create instances for each of the tuples used in the tests manually.

```scala

implicit def autoShowTuple[A: Show, B: Show]: Show[(A, B)] = {
  case (a, b) => s"(${a.show}, ${b.show})"
}
```

Add the ioPropF implicit def:
This function provides an implicit PropF instance for IO[A]. This instance lifts an IO[A] into an F[Expectations] by mapping the result of the IO to an Expectations value. In case of a failure, it creates an Expectations.Helpers.failure with the error message, and in case of success, it creates an Expectations.Helpers.success.

```scala
implicit def ioPropF[A]: PropF[IO[A]] = new Prop[F, IO[A]] {
  def lift(ioa: IO[A]): F[Expectations] = {
    ioa.attempt.map {
      case Left(e)  => Expectations.Helpers.failure(e.getMessage)
      case Right(_) => Expectations.Helpers.success
    }
  }
}
```

With these changes, you can now use the `forall` method without needing to create Show and PropF instances manually. The library will automatically derive the necessary instances when they are needed.

### DIFF
--- a/modules/scalacheck/src/weaver/scalacheck/Checkers.scala
+++ b/modules/scalacheck/src/weaver/scalacheck/Checkers.scala
@@ -22,6 +22,54 @@ trait Checkers {
   // Configuration for property-based tests
   def checkConfig: CheckConfig = CheckConfig.default
 
+  // Implicit defs for ioPropF
+  implicit def ioPropF[A]: PropF[F[A]] = new Prop[F, F[A]] {
+    def lift(fa: F[A]): F[Expectations] = {
+      attemptExpectation(fa.void)
+    }
+  }
+
+  // Add a new function to attempt the expectations
+  def attemptExpectation(fa: F[Unit]): F[Expectations] =
+    fa.attempt.map {
+      case Left(e)  => Expectations.Helpers.failure(e.getMessage)
+      case Right(_) => Expectations.Helpers.success
+    }
+
+// Implicit defs for Show
+  implicit def autoShowTuple[A: Show, B: Show]: Show[(A, B)] = {
+    case (a, b) => s"(${a.show}, ${b.show})"
+  }
+
+  implicit def autoShowTuple3[A: Show, B: Show, C: Show]: Show[(A, B, C)] = {
+    case (a, b, c) => s"(${a.show}, ${b.show}, ${c.show})"
+  }
+
+  implicit def autoShowTuple4[A: Show, B: Show, C: Show, D: Show]: Show[(A, B, C, D)] = {
+    case (a, b, c, d) => s"(${a.show}, ${b.show}, ${c.show}, ${d.show})"
+  }
+
+  implicit def autoShowTuple5[
+      A: Show,
+      B: Show,
+      C: Show,
+      D: Show,
+      E: Show]: Show[(A, B, C, D, E)] = {
+    case (a, b, c, d, e) =>
+      s"(${a.show}, ${b.show}, ${c.show}, ${d.show}, ${e.show})"
+  }
+
+  implicit def autoShowTuple6[
+      A: Show,
+      B: Show,
+      C: Show,
+      D: Show,
+      E: Show,
+      T: Show]: Show[(A, B, C, D, E, T)] = {
+    case (a, b, c, d, e, f) =>
+      s"(${a.show}, ${b.show}, ${c.show}, ${d.show}, ${e.show}, ${f.show})"
+  }
+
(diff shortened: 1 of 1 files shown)


modules/scalacheck/src/weaver/scalacheck/Checkers.scala
```diff
@@ -22,6 +22,54 @@ trait Checkers {
...
+  }
+
+  // Add a new function to attempt the expectations
+  def attemptExpectation(fa: F[Unit]): F[Expectations] =
+    fa.attempt.map {
+      case Left(e)  => Expectations.Helpers.failure(e.getMessage)
+      case Right(_) => Expectations.Helpers.success
+    }
+
+// Implicit defs for Show
```
Baccata, line 39: These are already provided by cats out of the box, they shouldn't be needed.

EstebanMarin: I agree, however, I had to manually implement the implicit where I am using the library, so I believe this can be abstracted in the library to make it easier to use, as the `implicit` is not evident to implement.
```
  test("For all sends 200's") {
    val testGenerator = for {
      x <- xGen
      y <- yGen
      z <- zGen
    } yield (x, y, z)
    forall(testGenerator) {
          case (x: X, y: Y, z: Z) =>
        Service[F]
          .IOMethod(x, y, z)
          .attempt
          .map {
            case Left(e: Throwable) => ???
            case Right(queue: Response) => ???
          }
    }
     implicit val xyzShow: Show[(x, y, z)] = Show.show {
    case (x, y, z) =>
      s"x: ${x}, y: ${y}, z: ${z}"
  }
    ```

Comment (Baccata) [reply modules/scalacheck/src/weaver/scalacheck/Checkers.scala:39]: > I had to manually implement the implicit where I am using the library

My point is that you shouldn't have : you automatically have Show instances for all tuple arities from 1 to 22 in scope. These instances should not be needed at all, and therefore I certainly won't accept their inlining in this class because. 

If you can provide a small workable reproduction (in the form of a scala-cli script or a small repository with an SBT or a mill build) that led you to think you needed the instances in question, I'd be happy to help you troubleshoot.

File: modules/scalacheck/src/weaver/scalacheck/Checkers.scala [hunk 2]
```diff
@@ -22,6 +22,54 @@ trait Checkers {
   // Configuration for property-based tests
   def checkConfig: CheckConfig = CheckConfig.default
 
+  // Implicit defs for ioPropF
+  implicit def ioPropF[A]: PropF[F[A]] = new Prop[F, F[A]] {
```
Baccata, line 26: I gave this some thoughts, and I don't think I'm confortable with this. It was a conscious decision to force the user to return expectations in their tests, in order to encourage them to separate the test logic from the expectations, and not intertwine them much. Allowing them to pass an arbitrary `F[A]` in `forall` calls would break cohesion with the rest of the library, and that's not really desirable. Especially because it's not the end of the world for people to add it in userland, exactly like you've done ... It's literally 10 lines of code. 

That being said, I can empathise with the fact that it's a annoying that there's not any short-hand way to turn an `F[A]` into an `F[Expectations]`. 

Maybe a better way would be to add an extension method that would allow for it. The code could look like : 

```scala  
forall(gen) { x  => 
   f(x).expectSuccess 
}
```

modules/scalacheck/src/weaver/scalacheck/Checkers.scala
EstebanMarin: Yes I agree with you here. I think I need to be more consistent on where I create the `Show` evidence.

modules/scalacheck/src/weaver/scalacheck/Checkers.scala
EstebanMarin: I that case I think, should close the PR. I did get better insights into the code. So thanks for that.

Issue #651, opened 2023-04-19
Fatal exceptions hang the runner


kubukoz:
Using 0.8.3, when you run this suite (or anything else that produces an exception of the "fatal" sort), it hangs the runner.

```scala
import weaver._

import cats.effect.IO

object DemoTestWeaver extends SimpleIOSuite {

  pureTest("demo") {
    throw new NoClassDefFoundError("fake exception")
  }
}
```

This is inconsistent with munit and scalatest's behavior, and hopefully it's something that can be fixed on weaver's side.

Related: #503

PR #653, opened 2023-04-24
Attempt to have fatal errors lead to the test process ending

Baccata:
Attempt at addressing https://github.com/disneystreaming/weaver-test/issues/651 

Just a quick POC to see whether something can be done.


Baccata:
> but in testOnly with just the weaver suite it still hangs. It can be cancelled with ctrl+C though. 

How did you test, out of curiosity ?

kubukoz:
oof I thought I'd linked the full repro in the original ticket...

https://github.com/kubukoz/demos/tree/sbt-weaver-hang

1. `test` (runs weaver, munit, scalatest)
2. `testOnly DemoTestWeaver`

PR #666, opened 2023-06-06
add documentation to suite types

mcanlas:
Wasn't sure if this was the canonical list of entrant suite types but it is a start. I believe these are the ones mentioned in the docs currently


Baccata (approved):
Thanks for this :)

PR #725, opened 2024-01-24
Grow output buffer when needed when reporting test outcomes (JS/Native)

### LINKED ISSUE #724: Failed IOs hang the runner on JS
```scala
package demo

import cats.effect.IO
import weaver._

object DemoTest extends SimpleIOSuite {

  test("this should fail rather than hang") {
    IO.unit *> IO.raiseError(new Throwable("woop woop")).as(success)
  }
}
```

This appears to hang on JS on both Scala 2.13.12 and 3.3.1.

Other versions involved:

- sbt: 1.9.8
- scala.js: 1.15.0
- weaver: 0.8.3

I will confirm whether this is a Weaver problem specifically or a regression in the other releases.

Full repro, with crossproject for comparison against the JVM: https://github.com/kubukoz/demos/tree/repro-js-fail-hang

Update: this seems to have started between 0.7.15 and 0.8.0. Diff: https://github.com/disneystreaming/weaver-test/compare/v0.6.15...v0.8.0

Update 2: bisected down to #563.

kubukoz:
Closes #724.

### DIFF
--- a/modules/framework/src-js-native/RunnerCompat.scala
+++ b/modules/framework/src-js-native/RunnerCompat.scala
@@ -15,6 +15,7 @@ import cats.syntax.all._
 import sbt.testing.{ EventHandler, Logger, Task, TaskDef }
 import java.io.ByteArrayOutputStream
 import java.io.DataOutputStream
+import java.util.Base64
 
 trait RunnerCompat[F[_]] { self: sbt.testing.Runner =>
   protected val args: Array[String]
@@ -27,15 +28,8 @@ trait RunnerCompat[F[_]] { self: sbt.testing.Runner =>
   private[weaver] val failedTests = ListBuffer.empty[(SuiteName, TestOutcome)]
 
   def reportDone(out: TestOutcomeNative): Unit = {
-    val serialised = ReadWriter.writer { p =>
-      p.writeString(out.suiteName)
-      p.writeString(out.testName)
-      p.writeDouble(out.durationMs)
-      p.writeString(out.verboseFormatting)
-      ()
-    }
     channel match {
-      case Some(send) => send(serialised)
+      case Some(send) => send(TestOutcomeNative.encode(out))
       case None       => failedTests.append(TestOutcomeNative.rehydrate(out))
     }
   }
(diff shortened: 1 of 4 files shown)


modules/framework/src-js-native/RunnerCompat.scala
```diff
@@ -205,12 +207,17 @@ private[weaver] object ReadWriter {
   }
 
   def writer(f: Writer => Unit): String = {
-    val buf = ByteBuffer.allocate(2048)
+    val baos = new ByteArrayOutputStream(2048)
+    val dos  = new DataOutputStream(baos)
```
kubukoz, line 211: **thought:** two words about this:

- 2048 is now the initial length of the array, but BAOS makes it grow when needed
- I tried to switch to `StringBuilder` as suggested, but the output should have constant-length encoding of numbers, and it seemed bothersome to try and pad them to make it work. Instead, `DataOutputStream` encodes the numbers in binary as expected.
- I'm not super comfortable with this encoding even in its current shape, because it assumes the bytes can be safely decoded from UTF-8 (`new String`) and encoded again (`.getBytes()`) without data loss, and I don't think that's generally true.

For example, this prints false:

```
val bb = ByteBuffer.allocate(20)
bb.putInt(42)
bb.putDouble(21.37)
val result = bb.array()
println(new String(result).getBytes().toList == result.toList)
```

so... I think ultimately the right solution would indeed be to use `StringBuilder`, albeit with a different encoding.

Thoughts?

kubukoz: ah, miraculously our doubles never have the floating point part (`.toMillis.toDouble`). However, even 50 000 ms (50 seconds) is enough to fail the roundrip.

kubukoz: ok, added some tests - 3a01e09bdf174b1ea909b9915e73bed15e98f590 - for the time being, encoding the bytes in base64 to bypass the UTF-8 problem, just to have a green build. Maybe it's fine.

kubukoz: finally, living proof of the roundtrip problem:

```scala
package demo

import cats.effect.IO
import weaver._
import concurrent.duration._

object DemoTest extends SimpleIOSuite {

  test("this should fail rather than hang") {
    IO.unit.as(failure("verbosity goes here"))
  }
}

object DemoTest2 extends SimpleIOSuite {

  test("this should fail rather than hang") {
    IO.unit.as(failure("verbosity goes here"))
  }
}
```

0.8.3:

image

another run:

image

206de20df299981816821c78f8fdf743b1c7540f :

image

Baccata (approved):

